PR #100, opened 2014-12-30
Add rake tasks for files added to spec.files by a cross_compile block

### LINKED ISSUE #98: Cannot add targets to spec.files after Rake initializes
While updating mysql2 to use rake-compiler 0.9.x for cross-compiling to Windows, I found that rake-compiler generates the file lib/mysql2/mysql2.rb with contents like
```
RUBY_VERSION =~ /(\d+.\d+)/
require "mysql2/#{$1}/mysql2"
```
But since this file isn't listed in the gemspec, packaging fails:
```
[snip]
install -c tmp/x86-mswin32-60/mysql2/2.0.0/mysql2.so lib/mysql2/2.0/mysql2.so
cp tmp/x86-mswin32-60/mysql2/2.0.0/mysql2.so tmp/x86-mswin32-60/stage/lib/mysql2/2.0/mysql2.so
** Execute native:mysql2:x86-mswin32-60
** Execute native:x86-mswin32-60
** Execute native
** Invoke gem (first_time)
** Invoke pkg/mysql2-0.3.17-x86-mingw32.gem (first_time)
** Invoke pkg (first_time, not_needed)
** Invoke pkg/mysql2-0.3.17-x86-mingw32 (first_time)
rake aborted!
Don't know how to build task 'tmp/x86-mingw32/stage/lib/mysql2/mysql2.rb'
```

I think this is happening since a3365c995770ceddba098937c2ec9a5e4b3914d9. In that commit, a rake task is created for each file to copy it to the target platform staging directory and then copy it back for packaging.

Either documentation or compile-time gemspec adjustment may be needed to include the generated file `lib/gemname/gemname.rb`.

sodabrew:
Resolves #98. Note that this code duplicates a block of code in `define_compile_tasks`. I could split it out to a private separate method.

### DIFF
--- a/lib/rake/extensiontask.rb
+++ b/lib/rake/extensiontask.rb
@@ -87,6 +87,26 @@ Rerun `rake` under MRI Ruby 1.8.x/1.9.x to cross/native compile.
     end
 
     private
+    # copy other gem files to staging directory
+    def define_staging_file_tasks(files, lib_path, stage_path, platf, ruby_ver)
+      files.each do |gem_file|
+        # ignore directories and the binary extension
+        next if File.directory?(gem_file) || gem_file == "#{lib_path}/#{binary(platf)}"
+        stage_file = "#{stage_path}/#{gem_file}"
+
+        # copy each file from base to stage directory
+        unless Rake::Task.task_defined?(stage_file) then
+          directory File.dirname(stage_file)
+          file stage_file => [File.dirname(stage_file), gem_file] do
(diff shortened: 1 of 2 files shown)


sodabrew:
Factored out a helper method, added a unit test, here's an example where I need this functionality: https://github.com/brianmario/mysql2/blob/master/tasks/compile.rake#L30

kou:
Thanks for adding a test!
I've merged it.

Issue #104, opened 2015-02-09
Changing cross_config_options for each cross_platform


sodabrew:
In brianmario/mysql2#473, I dreamed that the code below worked, but it doesn't. The issue I'm trying to solve is providing different options for each value in `ext.cross_platform`. It would be awesome to provide a hash like this:

``` ruby
    ext.cross_compile = true
    ext.cross_platform = ['x86-mingw32', 'x86-mswin32-60', 'x64-mingw32']
    ext.cross_config_options = {
      'x86-mingw32'    => [ "--with-mysql-dir=" + File.expand_path("../../vendor/#{vendor_mysql_dir('x86')}", __FILE__) ],
      'x86-mswin32-60' => [ "--with-mysql-dir=" + File.expand_path("../../vendor/#{vendor_mysql_dir('x86')}", __FILE__) ],
      'x64-mingw32'    => [ "--with-mysql-dir=" + File.expand_path("../../vendor/#{vendor_mysql_dir('x64')}", __FILE__) ],
    }
```

The code above results in this cross-compile configuration:
``` shell
cd tmp/x86-mingw32/mysql2/1.9.3
/home/vagrant/.rvm/rubies/ruby-1.9.3-p550/bin/ruby -I. ../../../../ext/mysql2/extconf.rb x86-mingw32 --with-mysql-dir=/vagrant/mysql2-sodabrew/vendor/mysql-connector-c-6.1.5-win32 x86-mswin32-60 --with-mysql-dir=/vagrant/mysql2-sodabrew/vendor/mysql-connector-c-6.1.5-win32 x64-mingw32 --with-mysql-dir=/vagrant/mysql2-sodabrew/vendor/mysql-connector-c-6.1.5-winx64
```

... which is obviously not gonna work.

Is there an existing way to do this? If not, are you open to the Hash idea above?

sodabrew:
Oh for goodness sake. Of course I figure out my bug 20 seconds after posting it.
``` diff
-    ext.cross_config_options = {
+    ext.cross_config_options << {
```
Now it works as advertised. I will submit a doc improvement to explain #74.

kou:
Or you can use

```ruby
ext.cross_config_options = [
  {
]

PR #105, opened 2015-02-09
Describe multiple cross platform options in the README

sodabrew:
I had the syntax slightly wrong and opened up #104. About a minute later I found my mistake. Here is a documentation update would have helped me understand how to use the new feature from #74.

Edit: also added an example of adding files to the gem package.

Sidenote: the docs seem pretty adamant about **i386**-mingw32, but I've been using **x86**-mingw32. Am I wrong, or are the docs out of date, or are the two variants synonyms?

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -284,9 +284,9 @@ on the Windows host system you're cross-compiling for. An example:
 
   # File: ~/.rake-compiler/config.yml
 
-  rbconfig-i386-mingw32-1.8.6: /path/to/ruby-1.8.6/rbconfig.rb
-  rbconfig-i386-mingw32-1.8.7: /path/to/ruby-1.8.7/rbconfig.rb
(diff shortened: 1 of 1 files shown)


kou:
Thanks!

> Sidenote: the docs seem pretty adamant about *i386*-mingw32, but I've been using *x86*-mingw32. Am I wrong, or are the docs out of date, or are the two variants synonyms?

*x86*-mingw32 should be used because i386 is normalized to x86 in RubyGems: https://github.com/rubygems/rubygems/blob/master/lib/rubygems/platform.rb#L66

rake-compiler should use *x86*.

sodabrew:
Ok I will update the PR to switch to x86 and to state the format of cross config options as an array of strings and hashes.

sodabrew:
Updated. Let me know if you like the combined example and doc update.

kou:
Thanks!!!
I've merged your pull request!!!

PR #107, opened 2015-03-16
Remove executable bits from lib source files

tjouan:
I don't think these lib source files are executable on purpose, so I
suggest removing executable bits.


kou:
It's reasonable.

PR #108, opened 2015-03-16
Improve git ignore config

tjouan:
* Remove patterns belonging to global git ignore config;
* Specify full paths where possible;
* Mark directories as such with a trailing slash.

Would you be interested by such a change? I think it's better for
occasional contributors so they wouldn't be surprised by unexpected
behavior caused by an entry existing while they think it can't.

I can update if my change in the present form is incorrect.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,7 @@
-.*.sw[pon]
-.ruby-version
-.rvmrc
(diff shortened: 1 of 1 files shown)


kou:
It's acceptable. I've merged it.

PR #109, opened 2015-03-16
Change make program detection with `make' fallback

tjouan:
On my platform (FreeBSD), make program is simply `make`, and is
provided in the base install. However, I have to run
`MAKE=make rake compile` to avoid this kind of error:

    cd tmp/amd64-freebsd10/uh/2.2.1
    rake aborted!
    Couldn't find a suitable `make` tool. Use `MAKE` env to set an alternative.
    /.../.gem/ruby/22/gems/rake-compiler-0.9.5/lib/rake/extensiontask.rb:465:in `make'
    /.../.gem/ruby/22/gems/rake-compiler-0.9.5/lib/rake/extensiontask.rb:155:in `block (2 levels) in define_compile_tasks'
    /.../.gem/ruby/22/gems/rake-compiler-0.9.5/lib/rake/extensiontask.rb:154:in `block in define_compile_tasks'
    Tasks: TOP => compile => compile:amd64-freebsd10 => compile:uh:amd64-freebsd10 => copy:uh:amd64-freebsd10:2.2.1 => tmp/amd64-freebsd10/uh/2.2.1/uh.so
    (See full trace by running task with --trace)
    zsh: exit 1

I'm not sure if the change I propose is acceptable, but I can work on
alternatives that would satisfy other requirements. I didn't found
some tests (cucumber or rspec) related to the original behavior
concerning make program detection, should I add some on the original
behavior before this change, and after update this one with new tests?

### DIFF
--- a/lib/rake/extensiontask.rb
+++ b/lib/rake/extensiontask.rb
@@ -6,6 +6,11 @@ require "rubygems/package_task"
 
 module Rake
   class ExtensionTask < BaseExtensionTask
+    MAKE_PROGRAM_DETECTION = {
+      'gmake -v'        => 'gmake',
+      'make -v'         => 'make',
+      'command -v make' => 'make'
+    }.freeze
 
     attr_accessor :config_script
     attr_accessor :cross_compile
@@ -453,9 +458,9 @@ Java extension should be preferred.
           if RUBY_PLATFORM =~ /mswin/ then
             'nmake'
           else
-            ENV['MAKE'] || %w[gmake make].find { |c|
-              system("#{c} -v >> #{dev_null} 2>&1")
-            }
+            ENV['MAKE'] || MAKE_PROGRAM_DETECTION.find { |test, program|
+              system("#{test} >> #{dev_null} 2>&1")
+            }[1]
           end
       end
 


kou:
Could you show output of `ruby -e 'p system("gmake -v")'` on your environment?

tjouan:
$ ruby -e 'p system("gmake -v")'
    nil

And the make test give this:

    $ ruby -e 'p system("make -v")' 
    usage: make [-BeikNnqrstWwX] 
                [-C directory] [-D variable] [-d flags] [-f makefile]
                [-I directory] [-J private] [-j max_jobs] [-m directory] [-T file]
                [-V variable] [variable=value] [target ...]
    false

---

I forgot to mention more details are included in commit message:

    Change make program detection with `make' fallback

      Some platforms provide a working `make' program out of the box, but
    some versions of this program won't return a successful exit status
    code when given `-v' as argument.
      One such platform is FreeBSD, providing the `make' program in the base
    system which is based on `pmake'. This version won't accept `-v':

    $ make -v
    usage: make [-BeikNnqrstWwX]
                [-C directory] [-D variable] [-d flags] [-f makefile]
                [-I directory] [-J private] [-j max_jobs] [-m directory] [-T file]
                [-V variable] [variable=value] [target ...]
    zsh: exit 2     make -v

      This change ensures detection with `gmake -v` or `make -v` will
    continue to work as expected, but adds a fallback on plain `make' when
    tested successfully with the `command' shell builtin.

kou:
Thanks. I understand.
How about finding an executable file from `ENV["PATH"].split(File::PATH_SEPARATOR)` instead of using `command`?

tjouan:
Is it something like this you had in mind?
tjouan/rake-compiler@17ff794c1acf60ce43de0121cbf30ab7a119d0ea

I did not checked specifically if it's a file, I'm not sure it's
necessary. I can update the commit with `File.file?` test if needed.

Regarding the PR I can push the new change here if you think it's ok,
if not please let me know how to improve it.

kou:
Oh...
I wrongly merged the original changes instead of tjouan/rake-compiler@17ff794...

Anyway, thanks for your report!

tjouan:
I prefer the behavior you merged, more efficient to remove tests with
real executions. My change was more conservative as I don't know the
full project very well :-)

Thanks!

PR #110, opened 2015-03-25
Update cucumber POSIX step for *BSD platforms

tjouan:
Cucumber test suite includes a "I'm running a POSIX operating system"
step. This change updates the host OS test to support most BSD
platforms, it should work with: FreeBSD, NetBSD, OpenBSD,
DragonFly BSD and maybe others.

Popular ruby projects like listen uses the same kind of regular
expression pattern:
https://github.com/guard/listen/blob/master/Gemfile#L11

### DIFF
--- a/features/step_definitions/cross_compilation.rb
+++ b/features/step_definitions/cross_compilation.rb
@@ -1,6 +1,6 @@
-# Naive way of looking into platforms, please include others like FreeBSD?
+# Naive way of looking into platforms
 Given %r{^I'm running a POSIX operating system$} do
-  unless RbConfig::CONFIG['host_os'] =~ /linux|darwin/ then
+  unless RbConfig::CONFIG['host_os'] =~ /linux|darwin|bsd|dragonfly/ then
     raise Cucumber::Pending.new("You need a POSIX operating system, no cheating ;-)")
   end
 end


features/step_definitions/cross_compilation.rb
```diff
@@ -1,6 +1,6 @@
-# Naive way of looking into platforms, please include others like FreeBSD?
+# Naive way of looking into platforms
 Given %r{^I'm running a POSIX operating system$} do
-  unless RbConfig::CONFIG['host_os'] =~ /linux|darwin/ then
+  unless RbConfig::CONFIG['host_os'] =~ /linux|darwin|bsd|dragonfly/i then
```
kou: Is `i` needed?

tjouan: Honestly I'm not sure, on FreeBSD it is not needed:

    [2] pry(main)> RbConfig::CONFIG['host_os']
    => "freebsd10"

I think it has to do with DragonFly, I'm going to check and will
confirm in a few days.

tjouan:
I checked a few BSD, all reports a lowercase string. I did a fresh
install of last DragonFly BSD release, and it's also lowercase:

    $ /usr/local/bin/ruby22 -e "puts RbConfig::CONFIG['host_os']"
    dragonfly4

  I think we don't need the `i` flag, `RbConfig::CONFIG['host_os']"
is filled at ruby build time and set by autotools, my guess is it's
always lowercase for those OS but I'm not familiar with autotools.

  I updated the pull request.

kou:
Thanks for confirming it!
I've merged this pull request.

tjouan:
No problem, thank you for the merge!

PR #116, opened 2015-07-14
Replace rake-compiler-dev-box in favour of rake-compiler-dock

larskanis:
[Rake-compiler-dev-box](https://github.com/tjschuck/rake-compiler-dev-box) is no longer maintained by it's author. Although there are several forks, each adding changes for a given target project, it is generally more extensive to set up and to integrate. Vagrant images are built on the users computer and are not versioned. Moreover the vagrant images are mutable, so that every build can make persistent changes to the box. In the end this makes each and every rake-compiler-dev-box rather unique.

[Rake-compiler-dock](https://github.com/larskanis/rake-compiler-dock) comes as a gem and solves these issues by using Docker instead of Vagrant. That makes rake-compiler-dock faster and more reliable. It also provides a simple API that let's the user integrate the cross build into the Rakefile of the given project. In addition, it is also usable on Windows.

Although JRuby was included in the first step, it didn't make much sense. JRuby can easily be installed native on each target platform. So now, rake-compiler-dock focuses on Windows binary gems only. This fits to the given README section, too.

To sum up, I wouldn't recommend rake-compiler-dev-box to users of rake-compiler, any longer.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -234,10 +234,11 @@ different 'host' OS's.
 
 ==== The Easy Way
 
-Use rake-compiler-dev-box, a virtual machine provisioned with all the necessary
-build tools.  With one command, you can cross-compile and package your gem into
-native, Java, and Windows fat binaries (with 1.8, 1.9, and 2.0 support).  See
-https://github.com/tjschuck/rake-compiler-dev-box for more information.
+Use rake-compiler-dock, a gem that makes use of a virtual machine provisioned with
+all the necessary build tools.  You can add a task to your Rakefile, that
+cross-compiles and packages your gem into Windows fat binaries (with 1.8 to 2.2
+and x86/x64 support). See https://github.com/larskanis/rake-compiler-dock for more
+information.
 
 ==== The Manual Way
 


kou:
Thanks for your suggestion. I've merged this pull request.

If you'll keep maintaining rake-compiler-dock, how about moving larskanis/rake-compiler-dock to rake-compiler/rake-compiler-dock? If we recommend rake-compiler-dock, it may be better. Users may feel relieved if rake-compiler-dock exists in the "official" organization.

larskanis:
@kou Good point - I not even thought about that! But as a companion project to rake-compiler, it makes sense to place it there. I'll keep maintaining rake-compiler-dock in the future, because I use it for several [open source  projects](https://github.com/larskanis/rake-compiler-dock/wiki/Projects-using-rake-compiler-dock) and for some closed ones. What do I have to do for the migration?

kou:
Thanks for accepting my suggestion.
I've invited you to "Owner" team of rake-compiler organization. The team has "admin" right.

Now, you can move larskanis/rake-compiler-dock to rake-compiler/rake-compiler-dock at https://github.com/larskanis/rake-compiler-dock/settings . You can find "Transfer ownership" item in the page.

larskanis:
Seems that everything worked. Thank you! Old links are redirected, now, but I'll update all places I know about to the new location. And thank you for taking over the maintenance of rake-compiler!

kou:
I also confirmed! :-)

PR #117, opened 2015-07-16
Update link to rake-compiler-dock due to the transfer to the rake-compiler organisation.

larskanis:


sodabrew:
This is awesome! Really excited to have these projects together in an org!

PR #118, opened 2015-08-27
Make customizable compiler Xlint option for JRuby native extension

cosmo0920:
While investigating native extension for JRuby, I noticed lack of functionality to specify Xlint option for `javac`.

Usually, deprecation warnings are suppressed as below.

As an example, msgpack-ruby's JRuby native extension compilation log:

``` log
...
警告: [options] ブートストラップ・クラスパスが-source 1.6と一緒に設定されていません
注意:ext/java/org/msgpack/jruby/ExtensionValue.javaは非推奨のAPIを使用またはオーバーライドしています。
注意:詳細は、-Xlint:deprecationオプションを指定して再コンパイルしてください。
警告1個
touch tmp/java/msgpack/.build
...
```

### DIFF
--- a/lib/rake/javaextensiontask.rb
+++ b/lib/rake/javaextensiontask.rb
@@ -15,6 +15,9 @@ module Rake
     # Generate class files for specific VM version
     attr_accessor :target_version
 
+    # Specify lint option
+    attr_accessor :lint_option
+
     def platform
       @platform ||= 'java'
     end
@@ -31,6 +34,7 @@ module Rake
       @debug          = false
       @source_version = '1.5'
       @target_version = '1.5'
+      @lint_option    = 'unchecked'
     end
 
     def define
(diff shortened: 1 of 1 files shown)


kou:
Why should we support customizing `-Xlint` option?
Is there any suitable value? `-Xlint:all`?

cosmo0920:
> Why should we support customizing -Xlint option?

ref: 
http://docs.oracle.com/javase/6/docs/technotes/tools/solaris/javac.html#nonstandard

> -Xlint
    Enable all recommended warnings. In this release, enabling all available warnings is recommended.

Official document says "enabling all available warnings is recommended", but I think always enable all available warnings is too verbose.

> > Is there any suitable value? -Xlint:all?

Hmm..., it's a bit difficult....
Ideally, default option is `-Xlint:all`. 

But, practically, `-Xlint:all` options is useless for restricted buffer terminal like cmd.exe.
Because flood of warnings delete old log that stored in buffer.
(cmd.exe's maximum buffer length is 9999, not infinite.)

So, I think suitable default Xlint option is `-Xlint:unchecked`.
Java 5 specification introduces Generics. This specification and compile time Type erasure origin bug is too hard debugging if `-Xlint:unchecked` option were not specified to javac.

kou:
> But, practically, -Xlint:all options is useless for restricted buffer terminal like cmd.exe.

I think that rake-compiler isn't used in cmd.exe. rake-compiler will be used on GNU/Linux, *BSD, OS X and so on.
Do you have any use case on Windows?

cosmo0920:
Ah, you are right.... rake-compiler will be used in non Windows environment.... :/

aqqaluk:
set it up

kou:
#158

PR #127, opened 2016-04-14
testing but....Add a gemspec

aqqaluk:
working but blind


kou:
Your approach isn't acceptable.
We should integrated with existing tasks. https://github.com/rake-compiler/rake-compiler/blob/master/tasks/gem.rake should be changed too.

PR #128, opened 2016-04-24
Use basename of ext name as target filename

mrkn:
As mkmf's create_makefile extracts the basename of the given extension name with directory prefixes, I want to make rake-compiler doing so.

PR #129, opened 2016-04-24
Fix install destination with directory prefixes

mrkn:
In addition to #128, the install destination should be fixed to make sure the directory prefixes are preserved.

For example, when the extension name is `foo/bar`, the target shared library `bar.so` will be installed in `lib` directory in the current master branch in spite of it should be in `lib/foo`.

Moreover, I've added some examples for the case with directory prefixes.


kou:
Thanks.
I've merged.

PR #134, opened 2016-05-08
Accept symbol name in extension task

k0kubun:
Since I write rake's normal task name in Symbol, I've used Symbol for the argument of `Rake::ExtensionTask.new` until rake-compiler v0.9.7. But rake-compiler v0.9.8 raises an error with such case in https://github.com/rake-compiler/rake-compiler/blob/v0.9.8/lib/rake/extensiontask.rb#L115.

I propose to accept Symbol for `Rake::ExtensionTask`'s name like `Rake::Task` https://github.com/ruby/rake/blob/v11.1.2/lib/rake/task.rb#L95.

### DIFF
--- a/lib/rake/baseextensiontask.rb
+++ b/lib/rake/baseextensiontask.rb
@@ -46,6 +46,7 @@ module Rake
 
     def define
       fail "Extension name must be provided." if @name.nil?
+      @name = @name.to_s
 
       define_compile_tasks
     end
--- a/spec/lib/rake/extensiontask_spec.rb
+++ b/spec/lib/rake/extensiontask_spec.rb
@@ -17,6 +17,11 @@ describe Rake::ExtensionTask do
         ext.name.should == 'extension_one'
       end
 
+      it 'should allow symbol as extension name assignation' do
+        ext = Rake::ExtensionTask.new(:extension_one)
+        ext.name.should == 'extension_one'
+      end
+
       it 'should allow string as extension name using block assignation' do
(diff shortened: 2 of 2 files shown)


kou:
Good catch!
Could you add a test?

lib/rake/baseextensiontask.rb
```diff
@@ -35,7 +35,7 @@ def initialize(name = nil, gem_spec = nil)
     end
 
     def init(name = nil, gem_spec = nil)
-      @name = name
+      @name = name.to_s
```
kou: This change works only for `Task.new(name)` case.
We need to care about `Task#name=` case.

How about stringifying in `#define`?

k0kubun:
> Could you add a test?

Added a test and squashed to https://github.com/rake-compiler/rake-compiler/pull/134/commits/af7dee631b4011b9f2255c8c83c3ee8ae984c374.

lib/rake/baseextensiontask.rb
k0kubun: > This change works only for Task.new(name) case.
> We need to care about Task#name= case.
>
> How about stringifying in #define?

Thank you for nice suggestion. Fixed to stringify name in `#define`. https://github.com/rake-compiler/rake-compiler/pull/134/commits/af7dee631b4011b9f2255c8c83c3ee8ae984c374

kou:
Thanks.
I've merged and released a new version.

PR #135, opened 2016-08-19
Fixing regular expression.

nicolasnoble:
The single `.` here matches anything. The regular expression was happily working but it's technically not what we want to do. We want to match an actual dot, not "any character".


kou:
Good catch!

PR #137, opened 2016-11-30
Set Ruby version constraints in the gemspec of cross gems.

larskanis:
This way the gem, that is not built for the ruby version in use, fails at install time rather than runtime.

Moreover a clear error message is given to the user or bundler can select the source gem instead.

### DIFF
--- a/lib/rake/extensiontask.rb
+++ b/lib/rake/extensiontask.rb
@@ -25,6 +25,7 @@ module Rake
       @cross_compiling = nil
       @no_native = false
       @config_includes = []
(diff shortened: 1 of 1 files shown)


kou:
Thanks.
I've released 1.0.3. Please try it.

larskanis:
Thanks for adding a test case for this! I didn't have an idea how to test this properly, but that looks good.

I'll give it a try and I'll add the latest rake-compiler version to rake-compiler-dock-0.6.0. It will bring a new base image (Ubuntu-16.10), updated compilers and cross build for Linux.

kou:
Great!

PR #138, opened 2017-03-17
Fix using wrong alt separator when cross compiling

aharpervc:
As stated in the commit, the alt separator should be based on host compilation platform, not cross compilation target platform. This fixes not being able to cross compile for windows, from windows, via intermediate linux build system (eg, using rake-compiler-dock).

Fixes https://github.com/rails-sqlserver/tiny_tds/issues/331

### DIFF
--- a/lib/rake/extensiontask.rb
+++ b/lib/rake/extensiontask.rb
@@ -533,7 +533,7 @@ Java extension should be preferred.
           RUBY_VERSION = "#{version}"
           RUBY_DESCRIPTION = "ruby \#{RUBY_VERSION} (\#{RUBY_RELEASE_DATE}) [\#{RUBY_PLATFORM}]"
         end
-        if RUBY_PLATFORM =~ /mswin|bccwin|mingw/
+        if '#{RUBY_PLATFORM}' =~ /mswin|bccwin|mingw/
           class File
             remove_const :ALT_SEPARATOR
             ALT_SEPARATOR = "\\\\"


aharpervc:
I'm not sure how to test this, other than that when targeting this branch, I can successfully build tiny tds via rake compiler dock with `bundle exec rake gem:windows`, whereas with the current release, the build fails, as I describe in https://github.com/rails-sqlserver/tiny_tds/issues/331

kou:
I couldn't reproduce it on Debian GNU/Linux sid. Why?

```text
% git clone git@github.com:rails-sqlserver/tiny_tds.git
% cd tiny_tds
% rake gem:windows
rake-compiler-dock bash -c "    bundle && rake cross native gem RUBY_CC_VERSION=2.4.0:2.3.0:2.2.2:2.1.6:2.0.0 CFLAGS=\"-Wall\"\n"
The git source `git://github.com/rake-compiler/rake-compiler-dock.git` uses the `git` protocol, which transmits data without encryption. Disable this warning with `bundle config git.allow_insecure true`, or switch to the `https` protocol to keep your data secure.
Fetching git://github.com/rake-compiler/rake-compiler-dock.git
Fetching gem metadata from https://rubygems.org/.............
Fetching version metadata from https://rubygems.org/.
Resolving dependencies...
Installing rake 10.5.0
Installing byebug 9.0.6 with native extensions
Installing connection_pool 2.2.1
Using mini_portile2 2.1.0
Using minitest 5.10.1
Using rake-compiler-dock 0.6.0 from git://github.com/rake-compiler/rake-compiler-dock.git (at master@da5ecc3)
Using bundler 1.13.7
Installing rake-compiler 0.9.5
Using tiny_tds 1.3.0 from source at `.`
Bundle complete! 7 Gemfile dependencies, 9 gems now installed.
Use `bundle show [gemname]` to see where a bundled gem is installed.
install -c tmp/x86-mingw32/tiny_tds/2.4.0/tiny_tds.so lib/tiny_tds/2.4/tiny_tds.so
cp tmp/x86-mingw32/tiny_tds/2.4.0/tiny_tds.so tmp/x86-mingw32/stage/lib/tiny_tds/2.4/tiny_tds.so
install -c tmp/x86-mingw32/tiny_tds/2.3.0/tiny_tds.so lib/tiny_tds/2.3/tiny_tds.so
cp tmp/x86-mingw32/tiny_tds/2.3.0/tiny_tds.so tmp/x86-mingw32/stage/lib/tiny_tds/2.3/tiny_tds.so
install -c tmp/x86-mingw32/tiny_tds/2.2.2/tiny_tds.so lib/tiny_tds/2.2/tiny_tds.so
cp tmp/x86-mingw32/tiny_tds/2.2.2/tiny_tds.so tmp/x86-mingw32/stage/lib/tiny_tds/2.2/tiny_tds.so
install -c tmp/x86-mingw32/tiny_tds/2.1.6/tiny_tds.so lib/tiny_tds/2.1/tiny_tds.so
cp tmp/x86-mingw32/tiny_tds/2.1.6/tiny_tds.so tmp/x86-mingw32/stage/lib/tiny_tds/2.1/tiny_tds.so
install -c tmp/x86-mingw32/tiny_tds/2.0.0/tiny_tds.so lib/tiny_tds/2.0/tiny_tds.so
cp tmp/x86-mingw32/tiny_tds/2.0.0/tiny_tds.so tmp/x86-mingw32/stage/lib/tiny_tds/2.0/tiny_tds.so
install -c tmp/x64-mingw32/tiny_tds/2.4.0/tiny_tds.so lib/tiny_tds/2.4/tiny_tds.so
cp tmp/x64-mingw32/tiny_tds/2.4.0/tiny_tds.so tmp/x64-mingw32/stage/lib/tiny_tds/2.4/tiny_tds.so
install -c tmp/x64-mingw32/tiny_tds/2.3.0/tiny_tds.so lib/tiny_tds/2.3/tiny_tds.so
cp tmp/x64-mingw32/tiny_tds/2.3.0/tiny_tds.so tmp/x64-mingw32/stage/lib/tiny_tds/2.3/tiny_tds.so
install -c tmp/x64-mingw32/tiny_tds/2.2.2/tiny_tds.so lib/tiny_tds/2.2/tiny_tds.so
cp tmp/x64-mingw32/tiny_tds/2.2.2/tiny_tds.so tmp/x64-mingw32/stage/lib/tiny_tds/2.2/tiny_tds.so
install -c tmp/x64-mingw32/tiny_tds/2.1.6/tiny_tds.so lib/tiny_tds/2.1/tiny_tds.so
cp tmp/x64-mingw32/tiny_tds/2.1.6/tiny_tds.so tmp/x64-mingw32/stage/lib/tiny_tds/2.1/tiny_tds.so
install -c tmp/x64-mingw32/tiny_tds/2.0.0/tiny_tds.so lib/tiny_tds/2.0/tiny_tds.so
cp tmp/x64-mingw32/tiny_tds/2.0.0/tiny_tds.so tmp/x64-mingw32/stage/lib/tiny_tds/2.0/tiny_tds.so
cd pkg/tiny_tds-1.3.0-x86-mingw32
mv tiny_tds-1.3.0-x86-mingw32.gem ..
cd -
cd pkg/tiny_tds-1.3.0-x64-mingw32
mv tiny_tds-1.3.0-x64-mingw32.gem ..
cd -
%
```

aharpervc:
Reproduce the failure of the original code, or reproduce the current ci test failure?

kou:
The failure of `rake gem:windows` as I showed.

aharpervc:
I see. The reason why you wouldn't see the failure on Linux is that the fix is cross compiling from Windows.

kou:
Umm... I don't understand why your change can fix the case.
Could you show both `fake.rb` with your change and without your change?

aharpervc:
See [here](https://github.com/rails-sqlserver/tiny_tds/issues/331) for context.

This pr results in the following change to fake.rb:

WITH PR:
Works; because Linux build host is what is tested in fake.rb for alt path separators.

```ruby
RUBY_PLATFORM = "x86-mingw32"

(snip)

if 'x86_64-linux' =~ /mswin|bccwin|mingw/
  class File
    remove_const :ALT_SEPARATOR
    ALT_SEPARATOR = "\\"
  end
end
```

WITHOUT PR:
Doesn't work; when cross compiling from Windows, this test incorrectly succeeds and tries to use the alt separator on a Linux build host machine.

```ruby
RUBY_PLATFORM = "x86-mingw32"

(snip)

if RUBY_PLATFORM =~ /mswin|bccwin|mingw/
  class File
    remove_const :ALT_SEPARATOR
    ALT_SEPARATOR = "\\"
  end
end
```

kou:
Thanks.

I think that `fake.rb` "without" this pull request has the same content both on Linux and cross compiling from Windows. Is it wrong?

Here is `tmp/x86-mingw32/tiny_tds/2.4.0/fake.rb` on Debian GNU/Linux:

```ruby
        # Pre-load resolver library before faking, in order to avoid error
        # "cannot load such file -- win32/resolv" when it is required later on.
        # See also: https://github.com/tjschuck/rake-compiler-dev-box/issues/5
        require 'resolv'

        class Object
          remove_const :RUBY_PLATFORM
          remove_const :RUBY_VERSION
          remove_const :RUBY_DESCRIPTION if defined?(RUBY_DESCRIPTION)
          RUBY_PLATFORM = "x86-mingw32"
          RUBY_VERSION = "2.4.0"
          RUBY_DESCRIPTION = "ruby #{RUBY_VERSION} (#{RUBY_RELEASE_DATE}) [#{RUBY_PLATFORM}]"
        end
        if RUBY_PLATFORM =~ /mswin|bccwin|mingw/
          class File
            remove_const :ALT_SEPARATOR
            ALT_SEPARATOR = "\\"
          end
        end

        posthook = proc do
          $ruby = "/usr/local/rvm/rubies/ruby-2.4.0/bin/ruby"
          untrace_var(:$ruby, posthook)
        end
        trace_var(:$ruby, posthook)
```

I could build gems on Debian GNU/Linux as I showed.

What is the difference of `fake.rb` "without" this pull request on Linux and cross compiling from Windows? If there is no difference, why is the problem occurred only "cross compiling from Windows" case"?

aharpervc:
I don't have debian installed, but I can test on macOS.

Also, there's a dimension to this which I did not clearly articulate (because these diagnostics are jumping through a half dozen abstraction layers)... I'm using rake-compiler-dock. Oops, probably should have lead with that. That's what's providing the linux build environment.

With & without PR for macOS is the same as as my previous comment (which is also the same output as your "without" `fake.rb` for debian).

> If there is no difference, why is the problem occurred only "cross compiling from Windows" case"?

That is the big question. From my linked diagnostics issue, this is what I came up with:

**Situation:**

Docker is booting and running `bundle && rake cross native gem RUBY_CC_VERSION=2.4.0 CFLAGS="-Wall"` inside the docker machine, and this process fails with the above error.

mkmf.log says:

```
"i686-w64-mingw32-gcc -E -I/usr/local/rake-compiler/ruby/i686-w64-mingw32/ruby-2.4.0/include/ruby-2.4.0/i386-mingw32 -I/usr/local/rake-compiler/ruby/i686-w64-mingw32/ruby-2.4.0/include/ruby-2.4.0/ruby/backward -I/usr/local/rake-compiler/ruby/i686-w64-mingw32/ruby-2.4.0/include/ruby-2.4.0 -I../../../../ext/tiny_tds -I/c/users/aharper/workspace/tiny_tds/ports/i686-w64-mingw32/include -D_WIN32_WINNT=0x0501 -D__MINGW_USE_VC2005_COMPAT -D_FILE_OFFSET_BITS=64  -s -O1 -fno-omit-frame-pointer -fno-fast-math -Wall  conftest.c -o conftest.i"
cc1: error: \c\users\aharper\workspace\tiny_tds\ports\i686-w64-mingw32\include: Invalid argument
checked program was:
/* begin */
1: #include "ruby.h"
2: 
3: #include <winsock2.h>
4: #include <windows.h>
5: #include <sybfront.h>
/* end */
```

I can confirm that `sybfront.h` **does** exist (along with `sybdb.h` and the other header files) in `ports/i686-w64-mingw32/include`. **It appears that the problem may be the directory slashes going the wrong way... somewhere.** Even though the include argument is `-I/c/users/aharper/workspace/tiny_tds/ports/i686-w64-mingw32/include`, which should be fine.

I grep'd for part of that path to see where it might be coming out with backslashes in but everywhere looks correct with forward slashes....... except in `fake.rb`, with `ALT_SEPARATOR`, based on `RUBY_PLATFORM`. So that is where I focused my diagnostics.

However, maybe this is incorrect? Maybe something somewhere else is keying off of `ALT_SEPARATOR`, after `fake.rb` is loaded in?

ladykraken (approved):

Franciscoleon69 (approved):

PR #140, opened 2017-04-20
Add more tests for gemspec mutation of native gems

larskanis:
While thinking about a different issue I noticed, that there are some mutations of the gemspec for native gems, which are currently untested. This PR adds proper tests for them.


kou:
Thanks!!!

PR #141, opened 2017-04-26
Regard C++ sources as source_files

k0kubun:
While we can write CRuby's extension in C++, `rake compile` does not hook `make` by `*.cc` file changes. When I develop C++ extension, I want to compile it incrementally by `rake compile`. 

So I changed rake-compiler to regard `*.cc` (and `*.cpp` for some users) as `source_files`.

### DIFF
--- a/lib/rake/extensiontask.rb
+++ b/lib/rake/extensiontask.rb
@@ -18,7 +18,7 @@ module Rake
     def init(name = nil, gem_spec = nil)
       super
       @config_script = 'extconf.rb'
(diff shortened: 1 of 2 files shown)


kou:
Thanks!

k0kubun:
When will it be available at rubygems.org? I always rely on this feature and it would be happy to have this without cloning GitHub repository on bundle install.

kou:
Done.

k0kubun:
Thank you!!

PR #148, opened 2018-08-28
Update source & target version defaults to 6

prashantvithani:
Source & Target version 5 are no longer supported for java8 and later. refer https://github.com/ruby-concurrency/concurrent-ruby/issues/747#issuecomment-416230078


kou:
Thanks.

Issue #149, opened 2018-12-07
Using 'libdir' path for jruby class path breaks cross-compilation


prashantvithani:
Reference: https://github.com/ruby-concurrency/concurrent-ruby/pull/778#issuecomment-444884137
https://github.com/rake-compiler/rake-compiler/pull/147#issuecomment-444829177

Caused by: #147

PR #151, opened 2019-01-02
Fix cross-platform compilation for Java extensions

prashantvithani:
Using `libdir` provided by `RbConfig` to compile java extension from MRI
breaks as it can't find `jruby.jar` under `lib` directory. Added a check
on file existence if the path is taken from RbConfig and if it doesn't
exist, check for the path which is set in environment JRUBY_HOME to find
correct `jruby.jar`.

Raise the error to abort compilation if `jruby.jar` is not found.

### DIFF
--- a/lib/rake/javaextensiontask.rb
+++ b/lib/rake/javaextensiontask.rb
@@ -213,15 +213,9 @@ execute the Rake compilation task using the JRuby interpreter.
         end
       end
 
-      unless jruby_cpath
-        libdir = RbConfig::CONFIG['libdir']
-        if libdir.start_with? "classpath:"
-          raise 'Cannot build with jruby-complete'
-        end
-        candidate = File.join(libdir, "jruby.jar")
-        jruby_cpath = candidate if File.exist? candidate
-      end
-
+      # jruby_cpath might not be present from Java-9 onwards as it removes
+      # sun.boot.class.path. Check if JRUBY_HOME is set as env variable and try
+      # to find jruby.jar under JRUBY_HOME
       unless jruby_cpath
         jruby_home = ENV['JRUBY_HOME']
         if jruby_home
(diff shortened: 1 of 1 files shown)


prashantvithani:
FYI: @pitr-ch

prashantvithani:
Fix #149

lib/rake/javaextensiontask.rb
```diff
@@ -212,13 +212,26 @@ def java_classpath_arg(*args)
         rescue => e
         end
       end
+
       unless jruby_cpath
         libdir = RbConfig::CONFIG['libdir']
         if libdir.start_with? "classpath:"
           raise 'Cannot build with jruby-complete'
         end
```
kou: Do we still need this condition?

lib/rake/javaextensiontask.rb
```diff
@@ -212,13 +212,26 @@ def java_classpath_arg(*args)
...
+
+      unless jruby_cpath
+        jruby_home = ENV['JRUBY_HOME']
+        if jruby_home
+          candidate = File.join(jruby_home, 'lib', 'jruby.jar')
+          jruby_cpath = candidate if File.exist? candidate
+        end
+      end
+
+      raise "jruby.jar path not found" unless jruby_cpath
```
kou: Can we show candidates we tried here? It'll be helpful for users who want to debug their environment.

lib/rake/javaextensiontask.rb
prashantvithani: Yep. While compiling with jruby-complete, the libdir is `uri:classloader:/META-INF/jruby.home/lib` which anyway doesn't contain `jruby.jar`. In this case, location of `jruby-complete.jar` is itself a classpath which is not set by default for Java 9 onwards as mentioned in the description of https://github.com/jruby/activerecord-jdbc-adapter/commit/995be701d58f58a34bc0d1448c9cd54f9fcd9e48.

lib/rake/javaextensiontask.rb
prashantvithani: Changed the error message

lib/rake/javaextensiontask.rb
kou: I see.

kou:
Thanks!

PR #154, opened 2019-06-12
Add uses_libruby setting for libruby-agnostic gems

### LINKED ISSUE #153: Prevent rake-compiler from overriding required_ruby_version?
I'm working on a gem that uses FFI and doesn't use libruby directly.

This means it doesn't need an `.so` file per Ruby version, and that the pre-compiled binary gems should have the same `required_ruby_version` as the original gemspec.

rake-compiler currently does this when cross-compiling:

https://github.com/rake-compiler/rake-compiler/blob/18b335a87000efe91db8997f586772150528f342/lib/rake/extensiontask.rb#L261-L264

Is there any way to override this behaviour?

glebm:
This is needed for gems that use the ffi gem instead of libruby.

Fixes #153


kou:
I don't want to add this option.
I want users to use `cross_compiling` block.

glebm:
With the advent of https://github.com/ffi/ffi, it should be supported out of the box, as it's easy to get it wrong (especially the `if multi && !uses_libruby` bit).

I've filed a similar request in Ruby itself for `mkmf`:
https://bugs.ruby-lang.org/issues/15914

PR #157, opened 2019-09-19
Add the -encoding argument optionally to java compiler

krypton:
Add the possibility to specify the -encoding argument do javac command.

This is useful for when you have multiple projects to compile and each one have different encodings on class files.

I know that we can set the JAVA_TOOL_OPTIONS env variable like this `ENV['JAVA_TOOL_OPTIONS'] = '-Dfile.encoding=windows-1252'` but if we do this it's gone affect all the JavaExtensionTask that we have defined to run.

### DIFF
--- a/lib/rake/javaextensiontask.rb
+++ b/lib/rake/javaextensiontask.rb
@@ -17,6 +17,8 @@ module Rake
     # Generate class files for specific VM version
     attr_accessor :target_version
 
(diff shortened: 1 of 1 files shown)


kou:
Thanks!

krypton:
@kou are you gonna publish this version to RubyGems?

kou:
Done.

Issue #156, opened 2019-07-26
NoMethodError: undefined method `split' for nil:NilClass in extensiontask.rb:515:in `ruby_api_version'


kputnam:
I'm not sure if I'm missing something obvious, but when running `rake native gem`... or even just `rake native`, it fails with

```
rake aborted!
NoMethodError: undefined method `split' for nil:NilClass
/Users/kputnam/.rvm/gems/ruby-2.6.3/gems/rake-compiler-1.0.7/lib/rake/extensiontask.rb:515:in `ruby_api_version'
/Users/kputnam/.rvm/gems/ruby-2.6.3/gems/rake-compiler-1.0.7/lib/rake/extensiontask.rb:262:in `block in define_native_tasks'
/Users/kputnam/.rvm/gems/ruby-2.6.3/gems/rake-12.3.3/lib/rake/task.rb:273:in `block in execute'
/Users/kputnam/.rvm/gems/ruby-2.6.3/gems/rake-12.3.3/lib/rake/task.rb:273:in `each'
/Users/kputnam/.rvm/gems/ruby-2.6.3/gems/rake-12.3.3/lib/rake/task.rb:273:in `execute'
/Users/kputnam/.rvm/gems/ruby-2.6.3/gems/rake-12.3.3/lib/rake/task.rb:214:in `block in invoke_with_call_chain'
/Users/kputnam/.rvm/rubies/ruby-2.6.3/lib/ruby/2.6.0/monitor.rb:230:in `mon_synchronize'
/Users/kputnam/.rvm/gems/ruby-2.6.3/gems/rake-12.3.3/lib/rake/task.rb:194:in `invoke_with_call_chain'
/Users/kputnam/.rvm/gems/ruby-2.6.3/gems/rake-12.3.3/lib/rake/task.rb:238:in `block in invoke_prerequisites'
/Users/kputnam/.rvm/gems/ruby-2.6.3/gems/rake-12.3.3/lib/rake/task.rb:236:in `each'
/Users/kputnam/.rvm/gems/ruby-2.6.3/gems/rake-12.3.3/lib/rake/task.rb:236:in `invoke_prerequisites'
/Users/kputnam/.rvm/gems/ruby-2.6.3/gems/rake-12.3.3/lib/rake/task.rb:213:in `block in invoke_with_call_chain'
/Users/kputnam/.rvm/rubies/ruby-2.6.3/lib/ruby/2.6.0/monitor.rb:230:in `mon_synchronize'
[...]

PR #160, opened 2019-11-12
Update default ruby version for native tasks

underscorgan:
When trying to build a native (non cross-compiled) gem I ran into #156.
To fix this, I updated the default for `ruby_versions` to `[ruby_ver]`
(which is set to `RUBY_VERSION`). This shouldn't affect cross-compiled
work since `ruby_versions` will be set to
`@ruby_versions_per_platform[platf]` when set.

### DIFF
--- a/lib/rake/extensiontask.rb
+++ b/lib/rake/extensiontask.rb
@@ -254,7 +254,7 @@ Java extension should be preferred.
           spec.platform = Gem::Platform.new(platf)
 
           # set ruby version constraints
-          ruby_versions = @ruby_versions_per_platform[platf] || []
+          ruby_versions = @ruby_versions_per_platform[platf] || [ruby_ver]
           sorted_ruby_versions = ruby_versions.sort_by do |ruby_version|
             ruby_version.split(".").collect(&:to_i)
           end


kou:
This can fix the exception but this generates a gem that only includes binary for Ruby that we use to run `rake native gem`. If we use Ruby 2.6, the generated gem only includes binary for Ruby 2.6. It doesn't usable with Ruby except Ruby 2.6 such as Ruby 2.4 and Ruby 2.5.

Is it useful?
Could you show your use case?

underscorgan:
Yes, that's useful for me. In this case I need to have a gem with native extensions installed on a docker container, but I don't want my container to have all the necessary build tools on it. So I'm building the gem in one stage of the build then copying it to the final container (with the same base image / ruby / etc).

If there's a better way to get compile the gem than this, happy to use that. This was the only thing I could make work.

kou:
For the use case, you can just copy `GEM_HOME`:

```dockerfile
FROM ubuntu:18.04 as build

RUN \
  apt update && \
  apt install -y \
    cmake \
    git \
    libssl-dev \
    make \
    pkg-config \
    ruby-dev

RUN gem install bundler
RUN git clone --recursive https://github.com/libgit2/rugged.git
WORKDIR rugged
RUN bundle install
RUN rake gem
RUN gem install pkg/*.gem

FROM ubuntu:18.04

RUN \
  apt update && \
  apt install -y \
    ruby

COPY --from=build /var/lib/gems/2.5.0/ /var/lib/gems/2.5.0/
```

PR #161, opened 2019-11-29
Fix missing Makefile.in in Ruby 2.7

qnighy:
Starting from Ruby 2.7.0.preview2 (ruby/ruby@c7477c3), Makefile.in is no
longer in the root. Instead it has been moved under ./template/.

This patch fixes the problem by allowing makefiles in both places.

Note that this doesn't fix all the problems building Ruby 2.7. It fails to generate `builtin_binary.inc` but I'm not sure the proper way to fix it.

### DIFF
--- a/tasks/bin/cross-ruby.rake
+++ b/tasks/bin/cross-ruby.rake
@@ -95,12 +95,16 @@ end
 
 # backup makefile.in
 file "#{USER_HOME}/sources/#{RUBY_CC_VERSION}/Makefile.in.bak" => ["#{USER_HOME}/sources/#{RUBY_CC_VERSION}"] do |t|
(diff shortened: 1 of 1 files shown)


kou:
Thanks for your pull request.
But reusing `"#{USER_HOME}/sources/#{RUBY_CC_VERSION}/Makefile.in.bak"` for `template/Makefile.in` is ad-hok. I've fixed this by using `".../template/Makefile.in.bak` for Ruby 2.7.

qnighy:
Confirmed to work. Thanks!

larskanis:
@knu Any chance to release 1.0.9? I'm just preparing rake-compiler-dock for ruby-2.7 and this patch is required.

kou:
Ah, sorry.
I've released it.

PR #162, opened 2019-12-22
Allow parallel builds of cross rubies

larskanis:
This is used in rake-compiler-dock: https://github.com/rake-compiler/rake-compiler-dock/blob/master/Dockerfile.mri#L85-L92 Without the `parallel` tool cross ruby builds take ages.

Download and extraction of sources can not run in parallel, since the sources are used for multiple targets. Therefore this part can now be separately executed by
```
  rake-compiler prepare-sources RUBY_CC_VERSION=...
```

Afterwards multiple cross rubies can be built in parallel by
```
  rake-compiler cross-ruby RUBY_CC_VERSION=... HOST=...
```

Access to `config.yml` must be synchronized, when running in parallel, which is done per File#flock .

### DIFF
--- a/tasks/bin/cross-ruby.rake
+++ b/tasks/bin/cross-ruby.rake
@@ -145,40 +145,42 @@ task :install => ["#{USER_HOME}/ruby/#{MINGW_HOST}/#{RUBY_CC_VERSION}/bin/ruby.e
 desc "Update rake-compiler list of installed Ruby versions"
 task 'update-config' do
   config_file = "#{USER_HOME}/config.yml"
-  if File.exist?(config_file) then
-    puts "Updating #{config_file}"
-    config = YAML.load_file(config_file)
-  else
-    puts "Generating #{config_file}"
-    config = {}
-  end
-
-  files = Dir.glob("#{USER_HOME}/ruby/*/*/**/rbconfig.rb").sort
-
-  files.each do |rbconfig|
-    version, platform = rbconfig.match(/.*-(\d.\d.\d).*\/([-\w]+)\/rbconfig/)[1,2]
-    platforms = [platform]
-
-    # fake alternate (binary compatible) i386-mswin32-60 platform
-    platform == "i386-mingw32" and
-      platforms.push "i386-mswin32-60"
-
-    platforms.each do |plat|
-      config["rbconfig-#{plat}-#{version}"] = rbconfig
-
-      # also store RubyGems-compatible version
-      gem_platform = Gem::Platform.new(plat)
-      config["rbconfig-#{gem_platform}-#{version}"] = rbconfig
+  File.open(config_file, File::RDWR|File::CREAT) do |config_fd|
+    config_fd.flock(File::LOCK_EX)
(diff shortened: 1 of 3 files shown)


tasks/bin/cross-ruby.rake
```diff
@@ -126,6 +126,8 @@ file makefile_in => [makefile_in_bak] do |t|
   }
 end
 
+task "prepare-sources" => makefile_in
```
kou: This approach isn't enough.
Parallel `rake-compiler prepare-sources` may create a broken prepared source:

```bash
rm -rf ~/.rake-compiler
for host in i686-w64-mingw32 x86_64-w64-mingw32; do
  ruby -Ilib bin/rake-compiler cross-ruby VERSION=2.6.0 HOST=$host &
done
wait
ls -lah ~/.rake-compiler/sources
# total 28M
# drwxr-xr-x  3 kou kou 4.0K Dec 25 06:27 .
# drwxr-xr-x  3 kou kou 4.0K Dec 25 06:27 ..
# drwxr-xr-x 27 kou kou 4.0K Dec 25 06:27 ruby-2.6.0
# -rw-r--r--  1 kou kou  14M Dec 25  2018 ruby-2.6.0.tar.bz2
# -rw-r--r--  1 kou kou  14M Dec 25  2018 ruby-2.6.0.tar.bz2.1
```

larskanis: Sure, in `rake-compiler-dock` I execute the task `prepare-sources`sequentially for all rubies. Afterwards `cross-ruby` is executed in parallel, which works since all source related tasks are already done and skipped.

I didn't try to run the download and extract part in parallel. Maybe locking the source directory or a file inside would work. Is this something I should investigate and you would consider merging?

larskanis:
The travis CI fails when running this PR, but I don't know why. The specs are running without any failures locally. I tried to reproduce the CI failures locally, using exactly the same versions of ruby and all gems, but it always succeeds on my end.

tasks/bin/cross-ruby.rake
kou: Ah, sorry. I misunderstood.

I'm positive for parallel build support. But I think that creating a lock file instead of using `flock` approach is better. (I didn't try this yet. So we may not be able to use lock file approach.)

kou: Or is it better that we add support for `rake-compiler cross-ruby VERSIONS=i686-w64-mingw32-2.6.0:x86_64-w64-mingw32-2.7.0` syntax and use Rake's parallel build feature?

larskanis: `rake-compiler cross-ruby VERSIONS=i686-w64-mingw32-2.6.0:x86_64-w64-mingw32-2.7.0` implemented per multitask sounds good.

Is there anything wrong in using `File#flock`? Why should we create a separate lock file instead? Locking the vulnerable file by the operation system is the cleanest approach IMHO.

larskanis:
I think that #169 is the better approach, so that I'll close this.

PR #164, opened 2019-12-26
Extend mingw search pattern to find x86_64 gcc

larskanis:
The previous pattern only recognized 32 bit compiler versions.

Cross compiler names are usually `i686-w64-mingw32-gcc` and `x86_64-w64-mingw32-gcc`.

Background is, that I'll probably split the single  docker image of rake-compiler-dock into target specific versions. And the `x64-mingw32` version currently fails, since it doesn't contain a recognized mingw compiler.


kou:
Thanks!

PR #167, opened 2019-12-31
Trigger compilation by 'cross' task, remove obsolete code and update README

larskanis:
So far cross build of gems was triggered by 'rake cross native gem'. The "cross" task adds the cross targets to the "native" task. Invoking of "native" then compiles for all cross targets and for the native running ruby. The "gem" task then packages everything into corresponding gem files.

This has the downside, that cross build always bundles native target and cross targets together. If only cross targets are wanted, then they have to be triggered directly by "native:<target>" tasks.

Rake-compiler-dock-1.0 uses a separate image for each target and it leads to issues, if the native target is triggered in addition to the image cross target. This commit therefore changes "cross" to trigger the cross targets without the native running ruby. The cross build of gems can now be done by:
```
    rake cross gem
```
Adding "native" to the command leads to the previous behaviour of building for cross and native rubies.

This commit also removes a bunch of code that predates to the logik before commit a3365c9 . To that time there wasn't a "stage" directory for cross rubies, so that the cross builds couldn't be proberly triggered. The "Sooo brittle" code removed the native running Linux or MacOS target from extensions that are not intended to be cross compiled for the native running ruby, but it was no generic solution. Getting rid of this code and trigger cross compile directly is far easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ as easy:
 
 Those **two** simple lines of code automatically added the Rake tasks needed to
 build your 'hello_world' extension. For example, checking the Rake tasks on
-MRI Ruby 1.8.x/1.9 returns something similar to:
+MRI Ruby returns something similar to:
 
     $ rake -T
     (in /home/user/my_extension)
@@ -251,7 +251,7 @@ several settings for `Rake::ExtensionTask`:
 ## Cross compilation - the future is now.
 
 Rake-compiler also provides a standardized way to generate, from either Linux
-or OSX, extensions and gem binaries for your Windows users!
+or OSX, extensions and gem binaries for your Windows and Linux users!
 
 How can this be you say? Simple, rake-compiler's cross compilation features
 take advantage of GCC's host/target capabilities to build 'target' binaries on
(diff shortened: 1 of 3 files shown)


larskanis:
I found a backward compatibility issue with this patch in rake-compiler-dock: FXRuby cross builds for x86-mingw32 and x64-mingw32 only. This is currently done per `rake cross native gem`. Due to [this line](https://github.com/rake-compiler/rake-compiler/blob/20795323f932e863b30a145f84a92466f1aba361/lib/rake/extensiontask.rb#L451) the native x86_64-linux build target is removed from the native prerequisites, so that only x86-mingw32 and x64-mingw32 targets are triggered by the above rake call. However with this patch applied, the x86_64-linux task is triggered in addition, but it fails, since the corresponding fox toolkit files aren't available for linux cross build.

So this would affect cross build done with rake-compiler-dock-0.7.x, when the rake-compiler gem is updated to a future (minor) version with this patch included. It's not an issue with rake-compiler-dock-1.0, since it's a major version change, with several incompatibilities.

So I think there are several options:
1. Ignore the incompatibility, since it affects only already released gems and no one will attempt to build fat binary gems without support for ruby-2.7 hence rake-compiler-dock-1.0.
2. Don't merge this patch and keep cross task as it is.
3. Change to patch to keep the line that removes the native build of the running platform.
4. Change the rake-compiler version at the major level.

For 2. and 3. the workaround to cross build target specific gems is to call `rake native:#{platform} gem` instead of `rake cross gem` as proposed in this patch.

larskanis:
I [removed this patch](https://github.com/rake-compiler/rake-compiler-dock/commit/fb341734a632dd730c34e999ed623f4e9aec4f36) in the final release of rake-compiler-dock-1.0.0 since it has regressions.

However I think there should be some simplification for [the proposed rake call](https://github.com/rake-compiler/rake-compiler-dock/blob/master/README.md#run-builds-in-parallel).
```sh
rake native:#{plat} pkg/#{exttask.gem_spec.full_name}-#{plat}.gem
```
This is not nice!

I think something like this should build the platform gem:
```sh
rake cross:#{plat}
```
or only
```sh
rake cross:available
```
which builds for the platform(s) available in the `~/.rake-compiler/config.yml`.

larskanis:
Better separation between cross and native tasks isn't possible without breaking compatibility. Therefore I'll close this PR.

PR #169, opened 2020-01-10
Allow building of cross rubies in parallel

larskanis:
Rubies can be build like so:
```
  rake-compiler cross-ruby VERSION=2.7.0:2.6.0 HOST=x86_64-w64-mingw32:i686-w64-mingw32
```

This builds the cross product of all ":" separated ruby and host versions. To force sequential builds add option "-j1".

This attempt of parallel cross ruby builds doesn't use file locking, but rake `multitask` only. It is an alternative way to #162 .

### DIFF
--- a/tasks/bin/cross-ruby.rake
+++ b/tasks/bin/cross-ruby.rake
@@ -41,106 +41,105 @@ end
 require 'rake/extensioncompiler'
 
 MAKE = ENV['MAKE'] || %w[gmake make].find { |c| system("#{c} -v > /dev/null 2>&1") }
-USER_HOME = File.expand_path("~/.rake-compiler")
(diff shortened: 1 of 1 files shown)


larskanis:
Year after year I prepare rake-compiler-dock for the new ruby version. So I rebased this PR to the current master. Any chance to get the parallel build patch merged? The diff is somewhat better to read when [ignoring whitespaces](https://github.com/rake-compiler/rake-compiler/pull/169/files?w=1).

larskanis:
Essentially the PR is about 
1. changing `chdir "path" do; end` to `sh "cmd", chdir: "path"`
2. changing all constants that are derived from the cross ruby version to variables
3. and making `:install` a multitask.

kou:
Sorry for late response...

PR #171, opened 2020-03-03
Call cross_compiling block for native platform too

larskanis:
Currently the cross_compiling block isn't called when cross compiling for the build platform.
This is due to the fact that define_native_tasks() is called first for the build platform
and later on for the cross platforms. The task "native:<gemname>:<platform>" is [defined
only once](https://github.com/rake-compiler/rake-compiler/blob/a86722f002290b1c7b7b0edae6cdd624618e8153/lib/rake/extensiontask.rb#L247) and therefore only without the callback. The callback is therefore not called
for the cross platform that equals the build platform.

OMHO the root issue is that rake-compiler doesn't has a clear separation between
native and cross tasks. However I think calling cross_compiling for any binary gem
(native or cross) is a suitable workaround.

This fixes bug https://github.com/sparklemotion/nokogiri/issues/1991#issuecomment-592755871

### DIFF
--- a/lib/rake/extensiontask.rb
+++ b/lib/rake/extensiontask.rb
@@ -234,7 +234,7 @@ Java extension should be preferred.
       end
     end
 
-    def define_native_tasks(for_platform = nil, ruby_ver = RUBY_VERSION, callback = nil)
+    def define_native_tasks(for_platform = nil, ruby_ver = RUBY_VERSION)
       platf = for_platform || platform
 
       # tmp_path
@@ -280,7 +280,7 @@ Java extension should be preferred.
           spec.files += ext_files
 
           # expose gem specification for customization
-          callback.call(spec) if callback
+          @cross_compiling.call(spec) if @cross_compiling
 
           # Generate a package for this gem
           pkg = Gem::PackageTask.new(spec) do |p|
@@ -425,7 +425,7 @@ Java extension should be preferred.
 
       # now define native tasks for cross compiled files
       if @gem_spec && @gem_spec.platform == 'ruby' then
-        define_native_tasks(for_platform, ruby_ver, @cross_compiling)
+        define_native_tasks(for_platform, ruby_ver)
       end
 
       # create cross task


kou:
Sorry for late response...
Could you show how to reproduce this with Nokogiri?

larskanis:
Build on x86_64-linux:

```ruby
$ git clone https://github.com/sparklemotion/nokogiri
$ cd nokogiri
$ git checkout 33e5b426f099da152904617e4e0be3c63df1f5dd~ # commit before the above fix
$ bundle
$ rake gem:linux
$ gem inst -l pkg/nokogiri-1.11.0.rc1-x86-linux.gem 
Nokogiri is built with the packaged libraries: libxml2-2.9.10, libxslt-1.1.34, zlib-1.2.11, libiconv-1.15.
Successfully installed nokogiri-1.11.0.rc1-x86-linux

$ gem inst -l pkg/nokogiri-1.11.0.rc1-x86_64-linux.gem 
Successfully installed nokogiri-1.11.0.rc1-x86_64-linux
```

Note the missing message `Nokogiri is built with the packaged libraries` of the `x86_64-linux`version. This is because the `cross_compiling` hook wasn't called.

flavorjones:
Hi, I'm still struggling with this issue in Nokogiri, see #2078 for context and help seeing the problem in the codebase.

I'd like to gently ask that this be merged, so that I can reliably build native gems for Nokogiri in the upcoming v1.11.0 release (that will coincide with Ruby 3). Thank you!

kou:
Sorry for my lazy response again...

I could reproduce this by https://github.com/rake-compiler/rake-compiler/pull/171#issuecomment-656843424 . Thanks.

The `cross_compiling` hook isn't called because it's not cross compiling. rake-compiler-dock uses x86_64-linux environment. So building gem for x86_64-linux isn't cross compiling.

How about using `Rake::ExtensionTask#no_native=`?

```diff
diff --git a/Rakefile b/Rakefile
index 6c740a90..7b9d5f44 100644
--- a/Rakefile
+++ b/Rakefile
@@ -128,6 +128,7 @@ else
   Rake::ExtensionTask.new("nokogiri", HOE.spec) do |ext|
     ext.lib_dir = File.join(*['lib', 'nokogiri', ENV['FAT_DIR']].compact)
     ext.config_options << ENV['EXTOPTS']
+    ext.no_native = (ENV["FORCE_CROSS_COMPILING"] == "true")
     ext.cross_compile  = true
     ext.cross_platform = CROSS_RUBIES.map(&:platform).uniq
     ext.cross_config_options << "--enable-cross-build"
diff --git a/tasks/cross-ruby.rb b/tasks/cross-ruby.rb
index b77d7b43..78cc6fd8 100644
--- a/tasks/cross-ruby.rb
+++ b/tasks/cross-ruby.rb
@@ -166,7 +166,7 @@ namespace "gem" do
       RakeCompilerDock.sh <<-EOT, platform: plat
         gem install bundler --no-document &&
         bundle &&
-        rake native:#{plat} pkg/#{HOE.spec.full_name}-#{plat}.gem MAKE='nice make -j`nproc`' RUBY_CC_VERSION=#{ENV["RUBY_CC_VERSION"]}
+        FORCE_CROSS_COMPILING=true rake native:#{plat} pkg/#{HOE.spec.full_name}-#{plat}.gem MAKE='nice make -j`nproc`' RUBY_CC_VERSION=#{ENV["RUBY_CC_VERSION"]}
       EOT
     end
   end
```

flavorjones:
@kou Thank you for responding!

I tried your suggestion, and :sparkles: it totally worked :sparkles:. It also addressed https://github.com/sparklemotion/nokogiri/issues/2076 which resulted in a "native" `nokogiri.so` being generated and packaged in addition to the "cross-compiled" libraries.

I'm going to play with this a bit more to make sure it's totally working the way I expect, but I think you've solved multiple problems for us. Thank you! :two_hearts:

kou:
Thanks for confirming it.
Can we close this?

flavorjones:
@kou If you'd like to close, OK. I'm pretty confident I can get https://github.com/sparklemotion/nokogiri/pull/2125 to pass CI.

Thank you again!

kou:
OK.
Closing.

larskanis:
@kou Thank you for the hint to use `no_native=true`! I added this to the rake-compiler-dock readme in https://github.com/rake-compiler/rake-compiler-dock/commit/8a0c7ceb13f7657333c40ae74c2e60669a2778fd .

What do you think about adding this directly to rake-compiler? The environment variable `RCD_HOST_RUBY_PLATFORM` is always [defined by](https://github.com/rake-compiler/rake-compiler-dock#environment-variables) rake-compiler-dock.

kou:
OK. You can use `RAKE_EXTENSION_TASK_NO_NATIVE=true` with 15f630adc4325f9cafb762e15a051c2ec8bd21e2.

larskanis:
Looks good - thanks!

PR #172, opened 2020-07-08
Default to 1.7-level bytecode in Java build

headius:
rake-compiler defaults to source level and output bytecode version 1.6, corresponding to Java 1.6.

JRuby has not supported 1.6 since the release of JRuby 9.0, five years ago, when we started requiring Java 7.

This PR updates the default versions to "1.7" level for Java 7.

All current supported versions of JRuby require Java 8, but I'm ok with using 1.7 here for any 9.0 stragglers.


kou:
Thanks!

PR #176, opened 2020-12-03
Add architecture ppc64le to travis build

ddeka2910:
Add support for architecture ppc64le.  
This is part of the Ubuntu distribution for ppc64le. This helps us simplify testing later when distributions are re-building and re-releasing. For more info tag @gerrith3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,7 @@
+os: linux
+arch:
+ - amd64
+ - ppc64le
 language: ruby
 notifications:
   webhooks:


kou:
Why do we need this?
Do we have architecture dependent code?

gerrith3:
@kou Can you prove that you don't?  ;)  Actually by testing on other architectures you do *prove* that you don't have architecture dependent code -or if you do that you still run on that target architecture.  Usually ruby projects are pretty arch independent but I can't tell about Rake.  True compilers are definitely not architecture independent but ported to each architecture.  @ddeka2910 already built your tree and verified that it does indeed work on ppc64le currently.  Ongoing testing will ensure that it never regresses, including verifying that none of its dependencies regress and that all components are available for it to run for people using ppc64le.

kou:
Did you confirm what is tested with our test suite?
Unfortunately, our test suite doesn't build any extension and assert the result. We have only mocked cases. It can't verify any architecture dependent behavior.

Do you want to improve our test suite?

kou:
We've moved to GitHub Actions from Travis CI because the Travis CI's new plan doesn't match us.

Issue #177, opened 2020-12-10
Fails on macOS: something wrong with LDFLAGS="-pipe -s"


qnighy:
Most probably due to https://github.com/rake-compiler/rake-compiler/pull/165. Changing `LDFLAGS=-pipe -s` to `LDFLAGS=-pipe` fixes the problem.

```
$ rake -f tasks/bin/cross-ruby.rake cross-ruby VERSION="2.7.2" HOST=x86_64-darwin11
mkdir -p /Users/qnighy/.rake-compiler/builds/x86_64-darwin11/ruby-2.7.2
cd /Users/qnighy/.rake-compiler/builds/x86_64-darwin11/ruby-2.7.2
/Users/qnighy/.rake-compiler/sources/ruby-2.7.2/configure --host=x86_64-darwin11 --target=x86_64-darwin11 --build=x86_64-apple-darwin19.4.0 --enable-shared --disable-install-doc --with-ext= LDFLAGS=-pipe -s --prefix=/Users/qnighy/.rake-compiler/ruby/x86_64-darwin11/ruby-2.7.2
checking for ruby... /Users/qnighy/.rbenv/versions/2.7.1/bin/ruby
tool/config.guess already exists
tool/config.sub already exists
checking build system type... x86_64-apple-darwin19.4.0
checking host system type... x86_64-pc-darwin11
checking target system type... x86_64-pc-darwin11
checking for x86_64-darwin11-clang... no
checking for x86_64-darwin11-gcc... no
checking for x86_64-darwin11-cc... no
checking for clang... clang
configure: WARNING: using cross tools not prefixed with host triplet
checking for x86_64-darwin11-gcc... clang
checking for x86_64-darwin11-gcc... (cached) clang
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables...
checking whether we are cross compiling... yes
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether clang accepts -g... yes
checking for clang option to accept ISO C89... none needed
checking for clang option to accept ISO C99... none needed
checking for x86_64-darwin11-clang++... no
checking for clang++... clang++
checking whether we are using the GNU C++ compiler... yes
checking whether clang++ accepts -g... yes
checking how to run the C preprocessor... clang -E
checking if minimum required OS X version is supported... yes
checking whether we are using the GNU C++ compiler... (cached) yes
checking whether clang++ accepts -g... (cached) yes
[...]

PR #178, opened 2020-12-10
Modify options for darwin platforms

qnighy:
It is known that we need a different set of options for the darwin platforms as seen in:

- https://github.com/protocolbuffers/protobuf/blob/v4.0.0-rc2/kokoro/release/ruby/macos/ruby/ruby_build_environment.sh#L27-L50
- https://github.com/sparklemotion/nokogiri/blob/v1.11.0.rc3/scripts/setup-osx-native-builders#L34-L39

This patch backports the modifications above, and also fixes https://github.com/rake-compiler/rake-compiler/issues/177.

### DIFF
--- a/tasks/bin/cross-ruby.rake
+++ b/tasks/bin/cross-ruby.rake
@@ -120,6 +120,23 @@ file "#{build_dir}/Makefile" => [build_dir, source_dir] do |t|
   # Force Winsock2 for Ruby 1.8, 1.9 defaults to it
   options << "--with-winsock2" if MAJOR == "1.8"
 
+  # We need a different set of options for darwin.
+  # - The "-s" option breaks /usr/bin/ld, so don't use it.
+  # - We want to statically link some libraries for portability.
+  if MINGW_TARGET =~ /darwin/
+    options = [
+      "--host=#{MINGW_HOST}",
+      "--target=#{MINGW_TARGET}",
+      "--build=#{RUBY_BUILD}",
+      '--enable-static',
+      '--disable-shared',
+      '--disable-install-doc',
+      '--without-gmp',
+      '--with-ext=',
+      'LDFLAGS=-pipe',
+    ]
+  end
+
   chdir File.dirname(t.name) do
     prefix = File.expand_path("../../../ruby/#{MINGW_HOST}/#{RUBY_CC_VERSION}")
     options << "--prefix=#{prefix}"


kou:
`cross-ruby.rake` is for building Windows binaries by Mingw-w64.
Why do you want to use it for macOS?

qnighy:
> `cross-ruby.rake` is for building Windows binaries by Mingw-w64.
> Why do you want to use it for macOS?

In fact, I'm not sure about the exact reason why `google-protobuf`, `grpc`, or `nokogiri` uses rake-compiler for non-Windows platforms. I made this patch only because I wanted to easily reproduce build environments for google-protobuf.

Nevertheless, I thought it's a common usecase because [rake-compiler-dock provides builds for Linux](https://github.com/rake-compiler/rake-compiler-dock/blob/5dacab741e04354107bc2fa51ed2b5f06aec422e/Rakefile#L11-L14).

kou:
> I wanted to easily reproduce build environments for google-protobuf.

How about using https://github.com/protocolbuffers/protobuf/blob/v4.0.0-rc2/kokoro/release/ruby/macos/ruby/ruby_build_environment.sh#L27-L50 instead of using raw rake-compiler?

> Nevertheless, I thought it's a common usecase because [rake-compiler-dock provides builds for Linux](https://github.com/rake-compiler/rake-compiler-dock/blob/5dacab741e04354107bc2fa51ed2b5f06aec422e/Rakefile#L11-L14).

I'm not sure about it's a common use case but `cross-ruby.rake` isn't for it.

qnighy:
> How about using https://github.com/protocolbuffers/protobuf/blob/v4.0.0-rc2/kokoro/release/ruby/macos/ruby/ruby_build_environment.sh#L27-L50 instead of using raw rake-compiler?

Yeah, that's what I'm doing now. And it took a while to figure out we need the following additional patch:

```diff
-    'LDFLAGS=-pipe -s',
+    'LDFLAGS=-pipe',
```

So I made this PR to share the knowledge with whoever needs it.

> I'm not sure about it's a common use case but `cross-ruby.rake` isn't for it.

I thought rake-compiler-dock shares philosophy with rake-compiler (because they're in the same org), but is it not?

kou:
Philosophy...?
I'm just the current maintainer not the original author but I don't have philosophy of this project... What do you want to say by it? "You should accept this patch"?
Anyway, we don't list the feature in our feature overview: https://github.com/rake-compiler/rake-compiler#feature-overview

If we add support for building different version Rubies for the current platform, we need to redesign and rename the current `cross-ruby.task`. It's not "cross-compiling".

We can't accept this patch because this approach is too ad-hoc. `MINGW_TARGET =~ /darwin/` is strange. MinGW is "Minimalist GNU for Windows". It's not related to Darwin.

qnighy:
> What do you want to say by it? "You should accept this patch"?

Depends on the answer. It's fine if `rake-compiler` is totally for MinGW only and `rake-compiler-dock`'s Linux builds are only working by chance. Or if Linux is supported but darwin is out of support, then that would be fine too, although unfortunate to me. I wanted to clarify which of them are supported.

> If we add support for building different version Rubies for the current platform, we need to redesign and rename the current cross-ruby.task. It's not "cross-compiling".

I agree that the current use-case is not a cross-compilation. However, I think supporting darwin hosts or darwin targets does sometimes imply cross-compilation because darwin runs on several ISAs.

> We can't accept this patch because this approach is too ad-hoc.

I'm confused again. Then is there a chance such a change is accepted if `cross-ruby.rake` is refactored in a non-ad-hoc way? Like renaming `MINGW_*` into some other names and supporting both `.exe` and non-`.exe` files...

larskanis:
I'm currently adding support for MacOS to rake-compiler-dock. It mostly works already - I expect to have it ready when ruby-3.0.0 comes out. First I aim to support `x86_64-apple-darwin`, maybe later `arm64-apple-darwin` as cross targets.

PR #179, opened 2020-12-10
Use tar.gz instead of tar.bz2

qnighy:
Starting from 3.0.0-preview2, Ruby stopped shipping `tar.bz2` archives. This PR changes the archive format to `tar.gz` for better compatibility.


kou:
Thanks!

PR #180, opened 2020-12-10
Upstream verconf patch from grpc/grpc

qnighy:
This patch upstreams @jtattermusch's https://github.com/grpc/grpc/pull/12779.

### DIFF
--- a/tasks/bin/cross-ruby.rake
+++ b/tasks/bin/cross-ruby.rake
@@ -130,6 +130,9 @@ end
 # make
 file "#{build_dir}/ruby.exe" => ["#{build_dir}/Makefile"] do |t|
   chdir File.dirname(t.prerequisites.first) do
+    # From https://github.com/grpc/grpc/issues/12161. Old rubies have race conditions around verconf.h.
+    sh "test -s verconf.h || rm -f verconf.h"  # if verconf.h has size 0, make sure it gets re-built by make
+
     sh MAKE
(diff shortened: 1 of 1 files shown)


tasks/bin/cross-ruby.rake
```diff
@@ -130,6 +130,9 @@ end
 # make
 file "#{build_dir}/ruby.exe" => ["#{build_dir}/Makefile"] do |t|
   chdir File.dirname(t.prerequisites.first) do
+    # From https://github.com/grpc/grpc/issues/12161. Old rubies have race conditions around verconf.h.
```
kou, line 133: Which versions are "Old rubies"?

qnighy: I looked into it as I didn't know the details. It turned out the "old rubies" here are 2.3.x or before. Added a comment about it.

qnighy: 2.3.0 is quite old to me, but protocolbuffers/protobuf still seems to provide builds for them https://github.com/protocolbuffers/protobuf/blob/v3.14.0/ruby/Rakefile#L84

kou: OK.
We don't support EOE-ed Rubies.

qnighy: I see. Thanks anyway for reviewing this!

PR #181, opened 2020-12-19
Update README to describe more recent ruby versions

larskanis:
The current README is quite outdated and should be updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ as easy:
 
 Those **two** simple lines of code automatically added the Rake tasks needed to
 build your 'hello_world' extension. For example, checking the Rake tasks on
-MRI Ruby 1.8.x/1.9 returns something similar to:
+MRI Ruby returns something similar to:
 
     $ rake -T
     (in /home/user/my_extension)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -251,7 +251,7 @@ several settings for `Rake::ExtensionTask`:
 ## Cross compilation - the future is now.
 
 Rake-compiler also provides a standardized way to generate, from either Linux
-or OSX, extensions and gem binaries for your Windows users!
+or OSX, extensions and gem binaries for your Windows and Linux users!
```
kou, line 254: I think that macOS users can't generate fat gems for Linux without rake-compiler-dock.

kou: Ah, Homebrew provides x86_64-elf-gcc.

How do users prepare Rubies for Linux on macOS?

larskanis: I think by specifying `rake-compiler cross-ruby HOST=x86_64-elf-gcc` or so. But I have little experience with MacOS only.

kou: `cross-ruby` expects only Mingw-w64 builds. For example, it uses `.exe` extension for Ruby binary.

rake-compiler doesn't support cross compiling fat gems for Linux. Could you revert related changes?

larskanis: > `cross-ruby` expects only Mingw-w64 builds.

Not by any means! It is pretty agnostic to the build target. Just the default is `mingw`, if no `HOST=triple` is given.

But you're right, `ruby.exe` is an [expected file target](https://github.com/rake-compiler/rake-compiler/blob/master/tasks/bin/cross-ruby.rake#L137). However this actually doesn't hurt - it's only that the file target is never satisfied so that the build is triggered in any case. Maybe we should fix this inaccuracy, now that `rake-compiler cross-ruby` is used for many different targets.

kou: It just works accidentally. It's not a "supported" feature. :-)

Could you revert this for now?

PR #182, opened 2020-12-20
Don't mask out build env vars for cross-ruby

larskanis:
This was introduced in commit a7c113e580c1565000166f9967da82ff2ecdaeaa to avoid interference with host build. However it prohibits setting important flags for cross-build.

Also don't force 'LDFLAGS=-pipe -s' in configure, which doesn't work with `clang` compiler targeting MacOS. It can now be set externally.

Fixes #178


kou:
OK.
Users should use rake-compiler carefully.

PR #184, opened 2021-01-25
IronRuby has been a dead project for 10 years

enebo:
Iron Ruby was abandoned as a project by Microsoft in spring of 2011.  It only supports Ruby 1.8.6 and I am wondering if rake-compiler will even load on that last release.  Time to remove this warning saying they do not support native c-exts?


kou:
Good catch!

PR #185, opened 2021-01-29
Avoid warning: assigned but unused variable - e

olleolleolle:
This PR avoids a Ruby warning.

Seen at https://travis-ci.org/github/ruby-concurrency/concurrent-ruby/jobs/755324784#L277


kou:
Good catch!

PR #189, opened 2021-03-26
Generate and copy *.dSYM debug information for macOS

dylanahsmith:
## Problem

I've opened a ruby PR to add support for generating *.dSYM debug information on macOS (https://github.com/ruby/ruby/pull/4310) but even with that, rake-compiler doesn't copy these *.dSYM to the lib directory so they will actually be used.  This means that there aren't any source filename and line numbers in the crash report during development of a native extension.

## Solution

To avoid depending on https://github.com/ruby/ruby/pull/4310, I've added a task to generate the *.dSYM directory if it isn't generated by ruby if `RbConfig::CONFIG['dsymutil']` is present.  I also extended the copy task to copy it into the stage and lib directories.

### DIFF
--- a/lib/rake/extensiontask.rb
+++ b/lib/rake/extensiontask.rb
@@ -132,14 +132,26 @@ module Rake
       directory lib_binary_dir_path
       directory stage_binary_dir_path
 
+      dsymutil = RbConfig::CONFIG['dsymutil']
+      if dsymutil.to_s.empty?
+        dsymutil = nil
+      else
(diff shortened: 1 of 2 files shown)


larskanis:
Given the changes to `mkmf.rb` in https://github.com/ruby/ruby/pull/4310 it looks like #191 would be an alternative approach to fix this issue. Of course that would only work with a ruby version containing https://github.com/ruby/ruby/pull/4310 .

dylanahsmith:
Leveraging `make install` makes sense.  I'll close this in favour of #191

PR #191, opened 2021-03-27
Use "make install" instead of manually copying extension binary

### LINKED ISSUE #190: Supplementary files are not copied to the lib directory
There is currently no way to install other files in addition to the extension binary. While there is a mechanism in mkmf to install files per `$INSTALLFILES`, there is no such thing in rake-compiler. I would like to make use of this feature [in the pg gem](https://github.com/ged/ruby-pg/blob/f6ccdae9db2375716d0d0dfd55e9cc54ca1cf141/ext/extconf.rb#L60-L62), but noticed, that it only works per `gem install` and not per `rake compile`. 

A possible solution is to make use of `make install` instead of copying the extension binary on our own. Another way would be to add a mechanism to rake-compiler that is similar to mkmf's `$INSTALLFILES`.

larskanis:
Until now rake-compiler uses `make` to build the extension binary, but doesn't use `make install` to copy it into the `lib` directory. This has the downside, that only the extension binary is copied, but no additional files. In particular files registered per mkmf's `$INSTALLFILES` mechanism are not respected. Using `make install` fixes this and possibly also #189.

With this PR, the call to `make install` is not invoked for cross ruby tasks but only for native compile. This changes the cross build, so that cross compiled extension binaries are no longer copied into the native `lib` directory. IMHO that makes sense in any case, since the cross compiled extension binaries belong into the platform gems only and are superfluous in the native `lib` directory. 

As another extension to this PR, it might be useful to run the cross packaging tasks through `make install` as well. However the cross build and packaging is way more complex and at least in case of the pg gem, it is not helpful, since the `$INSTALLFILES` are not necessary for cross builds.

Fixes #190

### DIFF
--- a/lib/rake/extensiontask.rb
+++ b/lib/rake/extensiontask.rb
@@ -115,6 +115,7 @@ module Rake
       tmp_path = "#{@tmp_dir}/#{platf}/#{@name}/#{ruby_ver}"
       stage_path = "#{@tmp_dir}/#{platf}/stage"
 
+      siteconf_path = "#{tmp_path}/.rake-compiler-siteconf.rb"
       tmp_binary_path = "#{tmp_path}/#{binary_path}"
       tmp_binary_dir_path = File.dirname(tmp_binary_path)
       stage_binary_path = "#{stage_path}/#{lib_path}/#{binary_path}"
@@ -132,10 +133,27 @@ module Rake
       directory lib_binary_dir_path
       directory stage_binary_dir_path
 
+      directory File.dirname(siteconf_path)
+      # Set paths for "make install" destinations
+      file siteconf_path => File.dirname(siteconf_path) do
+        File.open(siteconf_path, "w") do |siteconf|
+          siteconf.puts "require 'rbconfig'"
+          siteconf.puts "require 'mkmf'"
+          siteconf.puts "dest_path = mkintpath(#{File.expand_path(lib_path).dump})"
+          %w[sitearchdir sitelibdir].each do |dir|
+            siteconf.puts "RbConfig::MAKEFILE_CONFIG['#{dir}'] = dest_path"
+            siteconf.puts "RbConfig::CONFIG['#{dir}'] = dest_path"
+          end
+        end
+      end
+
       # copy binary from temporary location to final lib
       # tmp/extension_name/extension_name.{so,bundle} => lib/
-      task "copy:#{@name}:#{platf}:#{ruby_ver}" => [lib_binary_dir_path, tmp_binary_path] do
-        install tmp_binary_path, "#{lib_path}/#{binary_path}"
+      task "copy:#{@name}:#{platf}:#{ruby_ver}" => [lib_binary_dir_path, tmp_binary_path, "#{tmp_path}/Makefile"] do
+        # install in lib for native platform only
+        unless for_platform
+          sh "#{make} install", chdir: tmp_path
+        end
       end
       # copy binary from temporary location to staging directory
       task "copy:#{@name}:#{platf}:#{ruby_ver}" => [stage_binary_dir_path, tmp_binary_path] do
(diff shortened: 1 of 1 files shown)


lib/rake/extensiontask.rb
```diff
@@ -132,10 +133,27 @@ def define_compile_tasks(for_platform = nil, ruby_ver = RUBY_VERSION)
       directory lib_binary_dir_path
       directory stage_binary_dir_path
 
+      directory File.dirname(siteconf_path)
+      # Set paths for "make install" destinations
+      file siteconf_path => File.dirname(siteconf_path) do
+        File.open(siteconf_path, "w") do |siteconf|
+          siteconf.puts "require 'rbconfig'"
+          siteconf.puts "require 'mkmf'"
+          siteconf.puts "dest_path = mkintpath(#{File.expand_path(lib_path).dump})"
```
larskanis, line 142: This is the same mechanism as [used in rubygems](https://github.com/rubygems/rubygems/blob/17904c6312bf773e3e3b62589efd4e663f180c21/lib/rubygems/ext/ext_conf_builder.rb#L28-L34). However we're using an absolute `dest_path` here, which has to be passed through `mkintpath` to properly work on Windows. In contrast rubygems uses a relative path and subsequently copies the files from a temporary directory to the destination extension directory.

dylanahsmith (approved):
I also tested on MacOS with https://github.com/ruby/ruby/pull/4310 and confirmed that it does copy the `*.bundle.dSYM` file to lib

lib/rake/extensiontask.rb
```diff
@@ -132,10 +133,27 @@ def define_compile_tasks(for_platform = nil, ruby_ver = RUBY_VERSION)
...
+      end
+
       # copy binary from temporary location to final lib
       # tmp/extension_name/extension_name.{so,bundle} => lib/
-      task "copy:#{@name}:#{platf}:#{ruby_ver}" => [lib_binary_dir_path, tmp_binary_path] do
-        install tmp_binary_path, "#{lib_path}/#{binary_path}"
+      task "copy:#{@name}:#{platf}:#{ruby_ver}" => [lib_binary_dir_path, tmp_binary_path, "#{tmp_path}/Makefile"] do
+        # install in lib for native platform only
+        unless for_platform
+          sh "#{make} install", chdir: tmp_path
```
kou, line 155: Can we use `sh make, install, chdir: tmp_path` here?

larskanis: This is how I did it first, but then I noticed, that it doesn't work if one sets the `MAKE` environment variable to something like `MAKE=make -j8 V=1` . In this case a make binary with spaces in the name is tried to be executed instead of being interpret as command line options. In contrast build tools like autoconf or cmake accept options to make as part of the `MAKE` environment variable. I therefore think it's best how it is.

kou: Thanks.
I understand.

lib/rake/extensiontask.rb
```diff
@@ -132,10 +133,27 @@ def define_compile_tasks(for_platform = nil, ruby_ver = RUBY_VERSION)
...
+      directory File.dirname(siteconf_path)
+      # Set paths for "make install" destinations
+      file siteconf_path => File.dirname(siteconf_path) do
+        File.open(siteconf_path, "w") do |siteconf|
+          siteconf.puts "require 'rbconfig'"
+          siteconf.puts "require 'mkmf'"
+          siteconf.puts "dest_path = mkintpath(#{File.expand_path(lib_path).dump})"
+          %w[sitearchdir sitelibdir].each do |dir|
+            siteconf.puts "RbConfig::MAKEFILE_CONFIG['#{dir}'] = dest_path"
+            siteconf.puts "RbConfig::CONFIG['#{dir}'] = dest_path"
```
eregon, line 145: This is highly problematic, could you find a way which does not mutate RbConfig, as that can break many things?
See https://github.com/rake-compiler/rake-compiler/issues/202

PR #197, opened 2021-12-27
Fix other "lib binary paths"

konsolebox:
The fix in cd9daf3fd02345c303390f533b586376a45f2554 seems not enough since the .so file is still copied to an extra folder in the stage directory.  This file tasks created are also not right.

### DIFF
--- a/lib/rake/extensiontask.rb
+++ b/lib/rake/extensiontask.rb
@@ -82,9 +82,12 @@ module Rake
     private
     # copy other gem files to staging directory
     def define_staging_file_tasks(files, lib_path, stage_path, platf, ruby_ver)
+      # lib_binary_path
+      lib_binary_path = "#{lib_path}/#{File.basename(binary(platf))}"
+
       files.each do |gem_file|
         # ignore directories and the binary extension
-        next if File.directory?(gem_file) || gem_file == "#{lib_path}/#{binary(platf)}"
+        next if File.directory?(gem_file) || gem_file == lib_binary_path
         stage_file = "#{stage_path}/#{gem_file}"
 
         # copy each file from base to stage directory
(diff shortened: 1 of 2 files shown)


kou:
Could you show how to reproduce the problem?

konsolebox:
Using this [output in Appveyor](https://ci.appveyor.com/project/konsolebox/digest-kangarootwelve-ruby/builds/42009034/job/wjej7frwphmml3g5)  an example you can see in line 1635 the file is being copied to `tmp/i386-mingw32/stage/lib/digest/digest/kangarootwelve.so`. At some I found the so file existing both `lib/digest` and `lib/digest/digest`. I'm not sure how I was able to do it, but I've been playing with `lib/digest/kangarootwelve.so` being used as a file task dependency.

And of course I've also debugged the variables using `puts`.

konsolebox:
Ok so here's a way to demonstrate the issues:

```
git clone https://github.com/konsolebox/digest-xxhash-ruby.git
cd digest-xxhash-ruby
rake lib/digest/digest/xxhash.so
rake clobber
```
The command shows three things:

1) `tmp/x86_64-linux/digest/xxhash/3.0.2/digest/xxhash.so` is copied as `tmp/x86_64-linux/stage/lib/digest/digest/xxhash.so` instead of `tmp/x86_64-linux/stage/lib/digest/xxhash.so

2) `rake lib/digest/digest/xxhash.so` is invalid and shouldn't work but it did.

3) After running `rake clobber`, `lib/digest/xxhash.so` still exists because `rake-compiler` listed `lib/digest/digest/xxhash.so` as a clobber target instead.

Please ignore my earlier comment about `lib/digest/digest/<so_file>` existing.  It probably was caused by me mistakenly installing an older commit of rake-compiler while I was debugging the issue.

lib/rake/extensiontask.rb
```diff
@@ -250,6 +256,9 @@ def define_native_tasks(for_platform = nil, ruby_ver = RUBY_VERSION, callback =
       # lib_path
       lib_path = lib_dir
 
+      # lib_path
```
kou, line 259:
```suggestion
      # lib_binary_path
```

kou:
Thanks.
I understand.

konsolebox:
Thank you too. 👍

PR #166, opened 2019-12-27
Fix determining of ruby versions in "rake native gem"

larskanis:
"rake native gem" without "cross" didn't set the ruby version constraint.
Instead it failed with NoMethodError like so:
```
/ffi $ rake native gem
no configuration section for specified version of Ruby (rbconfig-i386-mingw32-2.6.3)
no configuration section for specified version of Ruby (rbconfig-x64-mingw32-2.6.3)
install -c build/x86_64-linux/ffi_c/2.6.3/ffi_c.so lib/ffi_c.so
cp build/x86_64-linux/ffi_c/2.6.3/ffi_c.so build/x86_64-linux/stage/lib/ffi_c.so
rake aborted!
NoMethodError: undefined method `split' for nil:NilClass
/home/lars/.rvm/gems/ruby-2.6.3/gems/rake-compiler-1.0.9/lib/rake/extensiontask.rb:515:in `ruby_api_version'
/home/lars/.rvm/gems/ruby-2.6.3/gems/rake-compiler-1.0.9/lib/rake/extensiontask.rb:262:in `block in define_native_tasks'
/home/lars/.rvm/gems/ruby-2.6.3/gems/rake-12.3.3/exe/rake:27:in `<top (required)>'
/home/lars/.rvm/gems/ruby-2.6.3/bin/ruby_executable_hooks:24:in `eval'
/home/lars/.rvm/gems/ruby-2.6.3/bin/ruby_executable_hooks:24:in `<main>'
Tasks: TOP => native => native:x86_64-linux => native:ffi:x86_64-linux
(See full trace by running task with --trace)
```

PR #199, opened 2022-01-12
Set correct `required_ruby_version` when some `RUBY_CC_VERSION`s are missing

### LINKED ISSUE #198: Set correct `required_ruby_version` when some `RUBY_CC_VERSION`s are missing
This was raised by @flavorjones in https://github.com/sparklemotion/nokogiri/pull/2409#issuecomment-1006056285 .

When `RUBY_CC_VERSION` has some ruby version that are not available as cross compile platforms, they are simply ignored for building the gem. In this case it just shows the warning `no configuration section for specified version of Ruby (rbconfig-x64-mingw32-2.3.3)` or similar. However it turned out, that the versions defined in `RUBY_CC_VERSION` are still respected in the `required_ruby_version` of the resulting gemspec, although no such binaries have been built.

The issue happens practically in rake-compiler-dock which has added the new platform "x64-mingw-ucrt" which supports only ruby-3.1. On the other hand the "x64-mingw32" doesn't support ruby-3.1, but only ruby-2.4 to 3.0. This is in accordance to the [change of RubyInstaller-3.1](https://rubyinstaller.org/2021/12/31/rubyinstaller-3.1.0-1-released.html). `RUBY_CC_VERSION` is set to `3.1.0:3.0.0:2.7.0:2.6.0:2.5.0:2.4.0` [in any case](https://github.com/rake-compiler/rake-compiler-dock/blob/71afb5f2ed400453a180f0220edeb59a15af299f/Dockerfile.mri.erb#L269). The resulting gems should reflect the version range of the given platform, but both platform gemspecs actually state ruby-2.4 to 3.1 compatibility, currently.

larskanis:
The `required_ruby_version` was set twice, although one in `define_native_tasks` is enough.

The original commit that introduced automatic setting of `required_ruby_version` was https://github.com/rake-compiler/rake-compiler/commit/0dc23504cb03ed2fb3c506e1bb58af48d3851d1e . It detected the ruby version entirely based on `RUBY_CC_VERSION`. Later on there was commit https://github.com/rake-compiler/rake-compiler/commit/0dc23504cb03ed2fb3c506e1bb58af48d3851d1e which introduced setting the `required_ruby_version` for native but not cross gems and only for the active ruby versions. This made the first detection obsolete, but it wasn't deleted so far. This is fixed by this PR.

Fixes #198

### DIFF
--- a/lib/rake/extensiontask.rb
+++ b/lib/rake/extensiontask.rb
@@ -377,9 +377,6 @@ Java extension should be preferred.
           @lib_dir = "#{@lib_dir}/#{$1}"
         end
 
-        # Update cross compiled platform/version combinations
-        @ruby_versions_per_platform[for_platform] << version
-
         define_cross_platform_tasks_with_version(for_platform, version)
 
         # restore lib_dir
--- a/spec/lib/rake/extensiontask_spec.rb
+++ b/spec/lib/rake/extensiontask_spec.rb
@@ -456,14 +456,17 @@ describe Rake::ExtensionTask do
 
       it "should set required_ruby_version from RUBY_CC_VERSION, set platform, clear extensions but keep metadata" do
         platforms = ["x86-mingw32", "x64-mingw32"]
-        ruby_cc_versions = ["1.8.6", "2.1.10", "2.2.6", "2.3.3", "2.10.1"]
+        ruby_cc_versions = ["1.8.6", "2.1.10", "2.2.6", "2.3.3", "2.10.1", "2.11.0"]
         ENV["RUBY_CC_VERSION"] = ruby_cc_versions.join(":")
         config = Hash.new
         ruby_cc_versions.each do |ruby_cc_version|
           platforms.each do |platform|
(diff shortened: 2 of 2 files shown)


lib/rake/extensiontask.rb
```diff
@@ -377,9 +377,6 @@ def define_cross_platform_tasks(for_platform)
           @lib_dir = "#{@lib_dir}/#{$1}"
         end
 
-        # Update cross compiled platform/version combinations
-        @ruby_versions_per_platform[for_platform] << version
```
kou, line 381: This line was added in #166 by you.
Could you confirm that this doesn't break #166 case?

larskanis: Originally it was added in #137, but modified later on and it was actually obsoleted by #166, but I didn't notice that.
I can confirm that both use cases (#137 and #166) work properly.

kou: Thanks.

flavorjones:
Thank you both!

PR #209, opened 2022-12-14
Add support for RubyGems 3.3.21 or later

flavorjones:
Since Rubygems 3.3.21, the Gem::Platform name always contains the library version for gnu platforms. So where the rake-compiler config entries previously were:

```yaml
---
rbconfig-x86_64-linux-gnu-2.7.0: "/usr/local/rake-compiler/ruby/x86_64-redhat-linux/ruby-2.7.0/lib/ruby/2.7.0/x86_64-linux-gnu/rbconfig.rb"
rbconfig-x86_64-linux-2.7.0: "/usr/local/rake-compiler/ruby/x86_64-redhat-linux/ruby-2.7.0/lib/ruby/2.7.0/x86_64-linux-gnu/rbconfig.rb"
```

with later versions of rubygems, it is only

```yaml
---
rbconfig-x86_64-linux-gnu-2.7.0: "/usr/local/rake-compiler/ruby/x86_64-redhat-linux/ruby-2.7.0/lib/ruby/2.7.0/x86_64-linux-gnu/rbconfig.rb"
```

This means that the current way of finding a matching runtime, by doing a string comparison on the config keys, is no longer appropriate. This is causing failing builds downstream in `rake-compiler-dock`.

This PR:

- extracts a new CompilerConfig class to encapsulate the logic
- uses `Gem::Platform#=~` to tell if the gem platform matches the runtime platform

### DIFF
--- a/lib/rake/compiler_config.rb
+++ b/lib/rake/compiler_config.rb
@@ -1,0 +1,12 @@
+module Rake
+  class CompilerConfig
+    def initialize(config_path)
+      require "yaml"
+      @config = YAML.load_file(config_path)
+    end
+
+    def find(ruby_version, gem_platform)
+      @config["rbconfig-#{gem_platform}-#{ruby_version}"]
+    end
+  end
+end
--- a/lib/rake/compiler_config.rb
+++ b/lib/rake/compiler_config.rb
@@ -6,7 +6,33 @@ module Rake
     end
 
     def find(ruby_version, gem_platform)
-      @config["rbconfig-#{gem_platform}-#{ruby_version}"]
+      gem_platform = Gem::Platform.new(gem_platform)
+
+      @config.each do |config_name, config_location|
+        # There are two variations we might find in the rake-compiler config.yml
+        #
+        # 1. config_name: rbconfig-x86_64-linux-3.0.0
+        #    runtime_platform_name: x86_64-linux
+        #    runtime_version: 3.0.0
+        #
+        # 2. config_name: rbconfig-x86_64-linux-gnu-3.0.0
+        #    runtime_platform_name: x86_64-linux-gnu
+        #    runtime_version: 3.0.0
+        #
+        # With rubygems < 3.3.21, both variations will be present (two entries pointing at the same
+        # installation).
+        #
+        # With rubygems >= 3.3.21, only the second variation will be present.
+        runtime_platform_name = config_name.split("-")[1..-2].join("-")
+        runtime_version = config_name.split("-").last
+        runtime_platform = Gem::Platform.new(runtime_platform_name)
+
+        if (ruby_version == runtime_version) && (gem_platform =~ runtime_platform)
+          return config_location
+        end
+      end
+
+      nil
     end
   end
 end
--- a/spec/lib/rake/compiler_config_spec.rb
+++ b/spec/lib/rake/compiler_config_spec.rb
@@ -27,4 +27,28 @@ describe Rake::CompilerConfig do
     expect(cc.find("2.7.0", "x86_64-linux")).to be_nil
     expect(cc.find("3.1.0", "arm64-linux")).to be_nil
   end
+
+  it "returns the matching config for inexact platform match" do
+    cc = Rake::CompilerConfig.new(config_file(<<~CONFIG))
+      ---
+      rbconfig-x86_64-linux-gnu-3.0.0: "/path/to/aaa/rbconfig.rb"
+      rbconfig-x86_64-linux-musl-3.1.0: "/path/to/bbb/rbconfig.rb"
+    CONFIG
+
+    expect(cc.find("3.0.0", "x86_64-linux")).to eq("/path/to/aaa/rbconfig.rb")
+    expect(cc.find("3.1.0", "x86_64-linux")).to eq("/path/to/bbb/rbconfig.rb")
+  end
+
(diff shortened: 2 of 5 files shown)


flavorjones:
@larskanis can you please review?

flavorjones:
I've force-pushed a fix to the test that was failing on some versions of Ruby, can someone please kick CI again? :pray:

lib/rake/compiler_config.rb
```diff
@@ -0,0 +1,23 @@
+require "yaml"
```
kou, line 1: We want to defer requiring `yaml`: https://github.com/rake-compiler/rake-compiler/commit/7e149899ff2f3ed9e73201405a1ffe106a72f978

lib/rake/compiler_config.rb
```diff
@@ -0,0 +1,23 @@
...
+    def initialize(config_path)
+      @config = YAML.load_file(config_path)
+    end
+
+    def find_matching_config(ruby_version, gem_platform)
+      # find a runtime platform that matches the gem platform
+      gem_platform = Gem::Platform.new(gem_platform)
+
+      config_key = @config.keys.find do |config_name|
+        runtime_platform_name = config_name.split("-")[1..-1][0..-2].join("-")
```
kou, line 14:
```suggestion
        runtime_platform_name = config_name.split("-")[1..-2].join("-")
```

kou, line 14: Could you add a comment to show sample values like the following?

```ruby
# With RubyGems < 3.3.21:
#  config_name = "rbconfig-x86_64-linux-3.0.0"
#   runtime_platform_name = "x86_64-linux"
#   runtime_version = "3.0.0"
# With RubyGems >= 3.3.21:
#   ...
```

lib/rake/compiler_config.rb
```diff
@@ -0,0 +1,23 @@
...
+  class CompilerConfig
+    def initialize(config_path)
+      @config = YAML.load_file(config_path)
+    end
+
+    def find_matching_config(ruby_version, gem_platform)
+      # find a runtime platform that matches the gem platform
+      gem_platform = Gem::Platform.new(gem_platform)
+
+      config_key = @config.keys.find do |config_name|
```
kou, line 13: How about the following style?

```ruby
@config.each do |name, config|
  # ...
  return config if found
end
nil
```

lib/rake/compiler_config.rb
```diff
@@ -0,0 +1,23 @@
+require "yaml"
+
+module Rake
+  class CompilerConfig
+    def initialize(config_path)
+      @config = YAML.load_file(config_path)
+    end
+
+    def find_matching_config(ruby_version, gem_platform)
```
kou, line 9: We can use simplify `find` here.

flavorjones:
@kou Thank you for the feedback! I've made those changes.

spec/lib/rake/compiler_config_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+describe Rake::CompilerConfig do
+  def config_file(contents)
+    Tempfile.new.tap do |tf|
+      tf.write(contents)
+      tf.close
+    end
+  end
+
+  it "returns the matching config for exact platform match" do
+    cc = Rake::CompilerConfig.new(config_file(<<~EOY))
```
kou, line 16: Could you use `CONFIG` or something that explains content instead of `EOX` style?

```suggestion
    cc = Rake::CompilerConfig.new(config_file(<<~CONFIG))
```

spec/lib/rake/compiler_config_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      rbconfig-x86_64-linux-gnu-3.0.0: "/path/to/aaa/rbconfig.rb"
+      rbconfig-x86_64-linux-musl-3.1.0: "/path/to/bbb/rbconfig.rb"
+    EOY
+
+    expect(cc.find("3.0.0", "x86_64-linux")).to eq("/path/to/aaa/rbconfig.rb")
+    expect(cc.find("3.1.0", "x86_64-linux")).to eq("/path/to/bbb/rbconfig.rb")
+  end
+
+  it "does not match the other way around" do
+    unless Gem::Requirement.new(">=3.3.21").satisfied_by?(Gem::Version.new(Gem::VERSION))
```
kou, line 43: How about simply using `Gem::Version#<`? 

```suggestion
    if Gem::Version.new(Gem::VERSION) < Gem::Version.new("3.3.21")
```

flavorjones: If you prefer that style, I'll make the change.

flavorjones:
@kou Thanks again for reviewing. I've addressed your suggestions! :heart:

kou:
Thanks!

flavorjones:
@kou I think we will need a new version, yes.

But there may be one more small change we'll need to make to support 3.2 in rake-compiler-dock (see https://github.com/rake-compiler/rake-compiler-dock/pull/82#issuecomment-1353454720). I'm still investigating that, though.

kou:
OK. Thanks for sharing the current status.

flavorjones:
@kou It looks like the other problem will be fixed upstream in Ruby. Please cut a release when you can. Thank you!

PR #211, opened 2023-03-06
baseextensiontask: provide an API to allow for dynamic sources

raggi:
If a build wishes to generate some of the extension source files then the pattern input is insufficient. Provide a place to stash an additional FileList that will then become a dependency of the compile tasks.

Example:

```ruby
file 'ext/libfoo/generated.c' do |t|
  open(t.name, 'w+) { |f| f << '#include <foo.h>' }
end

Rake::ExtensionTask.new('foo') do |ext|
  ext.extra_sources << 'generated.c'
end
```


kou:
Thanks. It makes sense.

PR #213, opened 2023-03-29
Don't use --release flag on Java 8

ahorek:
this allows using
```
  Rake::JavaExtensionTask.new("name", gemspec) do |ext|
    ext.release = '8'
  end
```
on Java 8 (for building the gem), because the flag is available since Java 9, see https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-source-and-target.html

this flag is for backward compatibility, so it's safe to just skip it if we can't use it.

relates to https://github.com/puma/puma/pull/3109 https://github.com/socketry/nio4r/pull/292

### DIFF
--- a/lib/rake/javaextensiontask.rb
+++ b/lib/rake/javaextensiontask.rb
@@ -212,7 +212,7 @@ execute the Rake compilation task using the JRuby interpreter.
     end
 
     def java_target_args
-      if @release
+      if @release && release_flag_supported?
         ["--release=#{@release}"]
       else
         ["-target", @target_version, "-source", @source_version]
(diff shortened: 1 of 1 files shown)


lib/rake/javaextensiontask.rb
```diff
@@ -303,5 +303,9 @@ def java_lint_arg
 
       "-Xlint:#{@lint_option}"
     end
+
+    def release_flag_supported?
+      !(RUBY_PLATFORM =~ /java/) || Gem::Version.new(Java::java.lang.System.getProperty('java.version')) >= Gem::Version.new("9")
```
kou, line 308: Could you use early return for readability?

```suggestion
      return false unless /java/match?(RUBY_PLATFORM)
      Gem::Version.new(Java::java.lang.System.getProperty('java.version')) >= Gem::Version.new("9")
```

ahorek: I guess I can't use ```match?``` because this gem is still ruby 1.8 compatible :) but sure, I'll change it...

kou: We can drop support for Ruby 1.8 because we can't test with it.

ahorek: ok, but it's for a separate PR, what should be the minimum version?

kou: I'm OK with a separate PR.

PR #214, opened 2023-03-29
drop ruby < 2.5 support

ahorek:
this PR drops support for long EOL rubies as suggested by https://github.com/rake-compiler/rake-compiler/pull/213#discussion_r1152473377

some gems still supports very versions, but I think Ruby 2.5 should be a reasonable minimum. This version is also currently tested by Github actions.
```
required_ruby_version = '>= 2.5.0'
```

it also applies rubocop autocorrect for Ruby 2.5 feature level (just styling)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,8 +1,8 @@
-source "https://rubygems.org"
+source 'https://rubygems.org'
 
-gem "rake"
+gem 'rake'
 
 group :development do
-  gem "rspec",    "~> 3.0"
-  gem "cucumber", "~> 2.0"
+  gem 'cucumber', '~> 2.0'
+  gem 'rspec',    '~> 3.0'
 end
--- a/bin/rake-compiler
+++ b/bin/rake-compiler
@@ -18,7 +18,7 @@ end
 Rake.application.init('rake-compiler')
 
 # Load the already cooked tasks ;-)
-load File.join(File.dirname(__FILE__), %w{.. tasks bin cross-ruby.rake})
+load File.join(File.dirname(__FILE__), %w[.. tasks bin cross-ruby.rake])
 
 # delegate control to Rake
 Rake.application.top_level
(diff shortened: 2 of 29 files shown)


konsolebox:
Please consider having 2.0 as minimum instead. I'm proud of having my gems support 2.2 even though the version is already EOL. I don't want to change dependency parameters just because of this.

Also it doesn't make sense to include style alterations in same commit.

Lastly, isn't using `=~` instead of `match?` good enough.

kou:
> it also applies rubocop autocorrect for Ruby 2.5 feature level (just styling)

Could you revert them? It adds many needles diffs.

> Please consider having 2.0 as minimum instead. I'm proud of having my gems support 2.2 even though the version is already EOL. I don't want to change dependency parameters just because of this.

Could you explain why do you want to support Ruby 2.2?
For example, I maintain test-unit and it supports old Rubies as much as possible (including Ruby 2.1: https://github.com/test-unit/test-unit/blob/master/.github/workflows/test.yml ) to support migration from an old Ruby to a recent Ruby. Upgrading Ruby itself and testing framework at the same time is difficult. So I want to support "old Ruby + old testing framework" -> "old Ruby + new testing framework" -> "new Ruby + new testing framework" migration path.

konsolebox:
> Could you explain why do you want to support Ruby 2.2?

2.2 is simply the minimum version my gems needed. They were first released on that version and I simply didn't have any reason to upgrade.

2.2 if I recall right made significant ext-related changes that made me decide to start with it instead of 2.0 or 2.1.

I don't even have compatibility hacks in my code and my 2.2 tests still succeed.

kou:
I see. Thanks.

ahorek:
@kou I understand the maintenance burden, there're still people using it for ancient stuff... if it doesn't break we shouldn't touch it without a strong reason...

kou:
I agree with you.
Note that we don't want to support EOL-ed Rubies actively for rake-compiler. If we need to create a shim for EOL-ed Rubies, we'll drop support for EOL-ed Rubies instead of creating a shim.

Issue #215, opened 2023-05-24
The extra_option including a space is splitted to execute the command.


junaruga:
The [ruby/openssl](https://github.com/ruby/openssl) is using the rake-compiler. Here is my test on Fedora 37.

```
$ ruby -v
ruby 3.3.0dev (2023-05-22T03:58:17Z master bd786e7896) [x86_64-linux]
```

```
$ git clone https://github.com/ruby/openssl.git

$ cd openssl
```

Install the gems. The `--standalone` option is my favorite. :)

```
$ bundle install --standalone

$ bundle list | grep rake-compiler
  * rake-compiler (1.2.1)
```

Then the following command with `--with-cflags='"-Wundef -Werror"'` propagates the `-Wundef -Werror` to the compiler gcc command. Note that the `MAKEFLAGS="V=1"` is an option to print the compiler commands.

```
$ MAKEFLAGS="V=1" bundle exec rake compile -- --enable-debug --with-cflags='"-Wundef -Werror"'
...
/home/jaruga/.local/ruby-bd786e7896/bin/ruby -I. -r.rake-compiler-siteconf.rb ../../../../ext/openssl/extconf.rb -- --enable-debug --with-cflags="-Wundef -Werror"
...
gcc -I. -I/home/jaruga/.local/ruby-bd786e7896/include/ruby-3.3.0+0/x86_64-linux -I/home/jaruga/.local/ruby-bd786e7896/include/ruby-3.3.0+0/ruby/backward -I/home/jaruga/.local/ruby-bd786e7896/include/ruby-3.3.0+0 -I../../../../ext/openssl  -DRUBY_EXTCONF_H=\"extconf.h\"    -fPIC -Wundef -Werror   -o openssl_missing.o -c ../../../../ext/openssl/openssl_missing.c
...
```

However, the following command with `--with-cflags='-Wundef -Werror'` (without additional quotes) separates the `--with-cflags=-Wundef` from the `-Werror`. And it doesn't propagate the `-Werror` to the gcc command line.

```
$ MAKEFLAGS="V=1" bundle exec rake compile -- --enable-debug --with-cflags='-Wundef -Werror'
...
/home/jaruga/.local/ruby-bd786e7896/bin/ruby -I. -r.rake-compiler-siteconf.rb ../../../../ext/openssl/extconf.rb -- --enable-debug --with-cflags=-Wundef -Werror
...
[...]

PR #217, opened 2023-05-29
handle nil in config options (broken in 1.2.2)

flavorjones:
__Context__

In rake-compiler 1.2.1 and earlier, if an extension did something like this, rake-compiler handled it gracefully:

```ruby
  Rake::ExtensionTask.new("gem", spec) do |ext|
    ext.config_options << ENV["EXTOPTS"] # this is probably going to be nil
  end
```

It is OK to do this because in 1.2.1 and earlier, the command array was executed as:

```
  sh cmd.join(' ')
```

However, this was changed in 1.2.2 to

```
  sh *cmd
```

which of course is preferable. However, this breaks extensions [like nokogiri that do something silly like above](https://github.com/sparklemotion/nokogiri/actions/runs/5109347425/jobs/9184057367).

__Details__

I've fixed Nokogiri (see https://github.com/sparklemotion/nokogiri/pull/2894) but I do think that there may be other gems impacted by this change.

This PR does a few things:

- extract a method `make_makefile_cmd` so that we can unit test it
- backfill basic test coverage
- backfill test for #215
- write a test for this `nil` case, and fix it by returning the result of `cmd.compact`

This is a big PR because of the extraction and test coverage. Please let me know if you would prefer this to be done a different way.

### DIFF
--- a/lib/rake/extensiontask.rb
+++ b/lib/rake/extensiontask.rb
@@ -79,6 +79,35 @@ module Rake
       end.flatten
     end
 
+    def make_makefile_cmd(root_path, tmp_path, extconf, siteconf_path, cross_platform) # :nodoc:
+      options = @config_options.dup
+
+      # include current directory
+      include_dirs = ['.'].concat(@config_includes).uniq.join(File::PATH_SEPARATOR)
+      cmd = [Gem.ruby, "-I#{include_dirs}", "-r#{File.basename(siteconf_path)}"]
+
+      # build a relative path to extconf script
+      abs_tmp_path = (Pathname.new(root_path) + tmp_path).realpath
+      abs_extconf = (Pathname.new(root_path) + extconf).realpath
+
+      # now add the extconf script
+      cmd << abs_extconf.relative_path_from(abs_tmp_path).to_s
+
+      if cross_platform
+        options.push(*cross_config_options(cross_platform))
+      end
+
+      # add options to command
+      cmd.push(*options)
+
+      # add any extra command line options
+      unless extra_options.empty?
+        cmd.push(*extra_options)
+      end
+
+      cmd
+    end
+
     private
     # copy other gem files to staging directory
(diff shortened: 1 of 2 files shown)


lib/rake/extensiontask.rb
```diff
@@ -79,6 +79,35 @@ def cross_config_options(for_platform=nil)
...
+        options.push(*cross_config_options(cross_platform))
+      end
+
+      # add options to command
+      cmd.push(*options)
+
+      # add any extra command line options
+      unless extra_options.empty?
+        cmd.push(*extra_options)
+      end
```
kou, line 106: How about simplifying this like others?

```suggestion
      cmd.push(*extra_options)
```

lib/rake/extensiontask.rb
```diff
@@ -188,34 +217,12 @@ def define_compile_tasks(for_platform = nil, ruby_ver = RUBY_VERSION)
...
-
-        # now add the extconf script
-        cmd << abs_extconf.relative_path_from(abs_tmp_path).to_s
-
-        # fake.rb will be present if we are cross compiling
-        if t.prerequisites.include?("#{tmp_path}/fake.rb") then
-          options.push(*cross_config_options(platf))
+        cross_platform = if t.prerequisites.include?("#{tmp_path}/fake.rb")
+          platf
         end
```
kou, line 222: Could you use the following style for readability?

```ruby
if ...
  cross_platform = platf
else
  cross_platform = nil
end
```

Or:

```ruby
cross_platform = nil
cross_platform = platf if ...
```

lib/rake/extensiontask.rb
flavorjones: I added a commit to clean up this method in a different way. I don't feel strongly and can do it as you requested if you prefer.

lib/rake/extensiontask.rb
flavorjones: Done!

kou:
Thanks!

PR #218, opened 2023-05-29
test: replace usage of `==` as an expectation with `eq` or `be`

flavorjones:
also make sure we run rspec with `--warnings` so we can see this more easily in the future.


kou:
Thanks!

PR #221, opened 2023-06-18
Add syntax-highlighting to example Ruby code in README.md

r7kamura:
I think that applying syntax highlighting here will provide a more readable README.

Since syntax highlighting cannot be performed on [indented code blocks](https://spec.commonmark.org/0.30/#indented-code-blocks), I replaced all ruby code parts with [fenced code blocks](https://spec.commonmark.org/0.30/#fenced-code-blocks) and specified `ruby` as their [info strings](https://spec.commonmark.org/0.30/#info-string). If you believe that the mixing of indented code blocks and fenced code blocks should be avoided, then I'm willing to rewrite all ones in fenced code blocks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,17 +69,19 @@ contribute back to your project.
 Now the fun part. It's time to introduce the code to your projects Rakefile
 to tell it to use rake-compiler to build your extension:
 
-    # File: extconf.rb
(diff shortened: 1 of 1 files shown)


kou:
Thanks!

Issue #202, opened 2022-02-02
Compile task modifies RbConfig values


bjfish:
The compile task currently modifies `RbConfig` values which should be avoided since these are intended to be read only and can cause issues for other users: https://github.com/rake-compiler/rake-compiler/blob/2b146623f239d490048bfe3cbe83785ac27ca040/lib/rake/extensiontask.rb#L148-L151

eregon:
For context, this caused a fairly complex issue in TruffleRuby: https://github.com/oracle/truffleruby/issues/2586
RubyGems uses `RbConfig::CONFIG['sitelibdir']` to find out the initial load path (https://github.com/rubygems/rubygems/blob/c98464e8e859157375177d2b88d9ee01c1355d3f/lib/rubygems.rb#L573-L581), and because that path is changed by rake compiler it doesn't find it anymore, and insert incorrectly in the $LOAD_PATH, loading incorrect gem versions.
This could happen on CRuby too with `--disable-gems` but luckily it does not thanks to CRuby setting the `@gem_prelude_index`, a pretty well hidden "hack" which saves CRuby here.
TruffleRuby will do the same for compatibility, but I believe no gem should mutate `RbConfig` as it can cause issue by anything else using it.

eregon:
This issue was introduced in https://github.com/rake-compiler/rake-compiler/commit/f48d669861583d6fb9c25accb03c1cc2a3292ea6, cc @larskanis

eregon:
In particular this breaks using Bundler (except the Bundler shipped with TruffleRuby, that works fine because there is then a single Bundler version) + the latest TruffleRuby release (22.0), and that's only fixable here.

kou:
@larskanis Can we use `make install sitearchdir=...` instead of rewriting `RbConfig::CONFIG["sitearchdir"]`?

larskanis:
As written in #191, rubygems is using the [same mechanism](https://github.com/rubygems/rubygems/blob/17904c6312bf773e3e3b62589efd4e663f180c21/lib/rubygems/ext/ext_conf_builder.rb#L28-L34) of changing `RbConfig` values when installing gems. That's why I didn't had a bad conscience proposing this.

> Can we use `make install sitearchdir=...` instead of rewriting `RbConfig::CONFIG["sitearchdir"]`?

I think so. `mkmf.rb` seems to not use the variables for anything but passing it into the Makefile. This could make the siteconf file obsolete.

eregon:
[...]

PR #222, opened 2023-07-27
Use ENV variables to set conf paths

ggmichaelgo:
closes: https://github.com/rake-compiler/rake-compiler/issues/202

Instead of creating `.rake-compiler-siteconf.rb` to override `RbConfig::MAKEFILE_CONFIG` and `RbConfig::CONFIG`, set environment variables to `make install` command. Since `RbConfig` is suppose to be readonly, we shouldn't overwrite its values.

Additionally, this update addresses a bug caused by `.rake-compiler-siteconf.rb`. The file's hardcoded destination path causes `rake-compiler` to consistently create files in that location, even if the project is renamed or moved. Consequently, changes to the gem project's path do not trigger the file task to update the file. This fix resolves this issue.

### DIFF
--- a/lib/rake/extensiontask.rb
+++ b/lib/rake/extensiontask.rb
@@ -1,7 +1,8 @@
 require "rbconfig"
+require "shellwords"
+
 require 'rake/baseextensiontask'
 require "rubygems/package_task"
-require "Shellwords"
 
 # Define a series of tasks to aid in the compilation of C extensions for
 # gem developer/creators.
--- a/lib/rake/extensiontask.rb
+++ b/lib/rake/extensiontask.rb
@@ -1,7 +1,7 @@
 require "rbconfig"
-
 require 'rake/baseextensiontask'
 require "rubygems/package_task"
+require "Shellwords"
 
 # Define a series of tasks to aid in the compilation of C extensions for
 # gem developer/creators.
@@ -79,7 +79,7 @@ module Rake
       end.flatten
     end
 
-    def make_makefile_cmd(root_path, tmp_path, extconf, siteconf_path, cross_platform) # :nodoc:
+    def make_makefile_cmd(root_path, tmp_path, extconf, cross_platform) # :nodoc:
       # include current directory
       include_dirs = ['.'].concat(@config_includes).uniq.join(File::PATH_SEPARATOR)
 
@@ -89,7 +89,7 @@ module Rake
       rel_extconf = abs_extconf.relative_path_from(abs_tmp_path).to_s
 
       # base command
-      cmd = [Gem.ruby, "-I#{include_dirs}", "-r#{File.basename(siteconf_path)}", rel_extconf]
+      cmd = [Gem.ruby, "-I#{include_dirs}", rel_extconf]
 
       # add all the options
       cmd += @config_options
@@ -140,7 +140,6 @@ module Rake
       tmp_path = "#{@tmp_dir}/#{platf}/#{@name}/#{ruby_ver}"
       stage_path = "#{@tmp_dir}/#{platf}/stage"
 
-      siteconf_path = "#{tmp_path}/.rake-compiler-siteconf.rb"
       tmp_binary_path = "#{tmp_path}/#{binary_path}"
       tmp_binary_dir_path = File.dirname(tmp_binary_path)
       stage_binary_path = "#{stage_path}/#{lib_binary_path}"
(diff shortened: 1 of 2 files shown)


kou:
Hmm... I think that users should run `rake clean` when they rename the project's folder.
If we always recreate `.rake-compiler-siteconf.rb`, we also always need to recreate all depended files such as `Makefile`. It will slow down `rake compile`.

ggmichaelgo:
Hello @kou !
I have found out about the issue https://github.com/rake-compiler/rake-compiler/issues/202, and I have read your https://github.com/rake-compiler/rake-compiler/issues/202#issuecomment-1046163347

I have updated the PR, and this will also solve my original issue :)

lib/rake/extensiontask.rb
```diff
@@ -158,26 +157,22 @@ def define_compile_tasks(for_platform = nil, ruby_ver = RUBY_VERSION)
...
-        end
-      end
-
       # copy binary from temporary location to final lib
       # tmp/extension_name/extension_name.{so,bundle} => lib/
       task "copy:#{@name}:#{platf}:#{ruby_ver}" => [lib_path, tmp_binary_path, "#{tmp_path}/Makefile"] do
         # install in lib for native platform only
         unless for_platform
-          sh "#{make} install target_prefix=", chdir: tmp_path
+          expaned_lib_path = mkintpath(File.expand_path(lib_path))
```
kou, line 165: Can we use relative path instead like RubyGems did https://github.com/rubygems/rubygems/pull/5626 ?

```suggestion
         relative_lib_path = Pathname(lib_path).relative_path_from(tmp_path)
```

I don't want to require `mkmf`.

lib/rake/extensiontask.rb
```diff
@@ -158,26 +157,22 @@ def define_compile_tasks(for_platform = nil, ruby_ver = RUBY_VERSION)
...
+
+          make_command = [
+            make,
+            "install",
+            "sitearchdir=#{expaned_lib_path}",
+            "sitelibdir=#{expaned_lib_path}",
+            "target_prefix=",
+          ]
+
+          sh(*make_command, chdir: tmp_path)
```
kou, line 175: We can't use this because `make` may have multiple shell components such as `make -j4`.

lib/rake/extensiontask.rb
ggmichaelgo: I have update the `sitearchdir` and `sitelibdir` to use the expanded path to be more debuggable :)

```ruby
expaned_lib_path = File.expand_path(lib_path)
```

ps. Thank you for reviewing the PR @kou !

lib/rake/extensiontask.rb
```diff
@@ -158,26 +156,20 @@ def define_compile_tasks(for_platform = nil, ruby_ver = RUBY_VERSION)
...
-        end
-      end
-
       # copy binary from temporary location to final lib
       # tmp/extension_name/extension_name.{so,bundle} => lib/
       task "copy:#{@name}:#{platf}:#{ruby_ver}" => [lib_path, tmp_binary_path, "#{tmp_path}/Makefile"] do
         # install in lib for native platform only
         unless for_platform
-          sh "#{make} install target_prefix=", chdir: tmp_path
+          expaned_lib_path = File.expand_path(lib_path)
```
kou, line 164: If we use absolute path, we need a special conversion on Windows that did by `mkintpath`:

https://github.com/ruby/ruby/blob/b5c74d548872388921402ff2db36be15e924a89b/lib/mkmf.rb#L1944-L1958

We don't need it when we use relative path.

kou: We also need to escape this path because this path may include a special character such as ` `.

Hmm. We may want to use `sh(*make_command_line, chdir: tmp_path)` with `Shellwords.shellsplit`:

```ruby
relative_lib_path = Pathname(lib_path).relative_path_from(tmp_path) 
make_command_line = Shellwords.shellsplit(make)
make_command_line << "install"
make_command_line << "sitearchdir=#{relative_lib_path}"
make_command_line << "sitelibdir=#{relative_lib_path}"
make_command_line << "target_prefix="
sh(*make_command_line, chdir: tmp_path)
```

ggmichaelgo: My initial thought of using relative path is that it would be easier to debug. However, the relative path is still debuggable and it will be better to avoid requiring `mkmkf`. (I just checked the `mkmkf`'s file size 😅 )

Also, thank you for pointing out the `Shellwords.shellsplit`, and it will be much more safer to use it!

kou:
Could you update the PR title and description before we merge this?
(We use the PR title and description for commit message. Could you also include `fix GH-202` to close #202 automatically when we merge this?)

kou:
Thanks!

Issue #225, opened 2023-08-02
Release 1.2.4 has issues with make.exe path on Windows


ls5302:
Building eventmachine 1.2.7 (with SSL) on a Windows 10 machine using Ruby 3.1.2 and rake-compiler v1.2.4 produces:

```text
Command failed with status (127): [C:msys32userbin/make.exe install sitearchdir...]

Tasks: TOP => compile => compile:i386-mingw32 => compiler:rubyeventmachine:i386-mingw32 => copy:rubyeventmachine:i386-mingw32:3.1.2
```

Using rake-compiler v1.2.3 works with no problems.

Looks like some backslashes have disappeared.

Probably the change in extensiontask.rb around line 168.

```text
Shellwords,shellsplit('c:\foo\bar\make.exe') => c:foobarmake.exe
Shellwords,shellsplit('c:/foo/bar/make.exe') => c:/foo/bar/make.exe
```

Using `set MAKE=c:/msys32/usr/bin/make.exe` appears to be a possible workaround as we use forward slashes.

kou:
Could you try 1.2.5?

ls5302:
All good in 1.2.5 - thank you.

PR #227, opened 2023-08-02
test: add windows and ruby 3.2 to the test matrix

flavorjones:
Trying to generate a failing test for #225


flavorjones:
Hmm. Got other failures, but not the issue reported at #225. Closing for now until I or someone else has time to spend more time on it.

PR #233, opened 2024-01-22
`require` the target `rbconfig` rather than copypasting verbatim

### LINKED ISSUE #232: Cross Compile looks for Ruby headers in the wrong directory
```
mkmf.rb can't find header files for ruby at <absolute/path/to/config/`tmp_dir`>/<`cross_platform`>/<`name`>/include/ruby.h

You might have to install separate package for the ruby development
environment, ruby-dev or ruby-devel for example.
```

Edit: It could be that I used my existing RubyIntaller2 prebuild instead of letting `rake-compiler` rebuild the entire Ruby from source.

### Analysis

Cross Compile currently works by replacing `RbConfig::CONFIG` with a `fake.rb` script in the `$LOAD_PATH`. This `fake.rb` copy-pastes the the target platform’s `rbconfig.rb` **verbatim**.
https://github.com/rake-compiler/rake-compiler/blob/9baeced144c20c3f40be2d678ac5b7f3ef4654a4/lib/rake/extensiontask.rb#L415-L435
However, `rbconfig.rb` begins with this for RubyIntaller2:
```ruby
# Ruby installed directory.
TOPDIR = File.dirname(__FILE__).chomp!("/lib/ruby/3.3.0/x64-mingw-ucrt")
```
Same for Ubuntu, for that matter:
```ruby
# Ruby installed directory.
TOPDIR = File.dirname(__FILE__).chomp!("/lib/ruby/3.3.0/x86_64-linux")
```

With `__FILE__` being `/path/to/project/tmp/fake.rb` rather than `/path/to/target/ruby/rbconfig.rb`, `RbConfig::TOPDIR` becomes `nil` and leads to a cascade of fallbacks.

ParadoxV5:
Fixes #232

### DIFF
--- a/lib/rake/extensiontask.rb
+++ b/lib/rake/extensiontask.rb
@@ -429,7 +429,7 @@ Java extension should be preferred.
           f.puts("require 'rbconfig'")
           f.puts("original_enable_shared = RbConfig::CONFIG['ENABLE_SHARED']")
           f.puts(fake_rb(for_platform, ruby_ver))
-          f.puts(File.read(t.prerequisites.first))
+          f.puts("require #{t.prerequisites.first.dump}")
           f.puts("RbConfig::CONFIG['ENABLE_SHARED'] = original_enable_shared")
         end
       end


ParadoxV5:
Greens CI: https://github.com/ParadoxV5/rake-compiler/actions/runs/7617624734

ParadoxV5:
The substitute `mkmf.rb` is also copy-pasted but possibly with replacements (not verbatim).
It allows us to continue supporting *Ruby 1.8 – **or do we?***

kou:
I'll merge this but we don't mark your usage https://github.com/rake-compiler/rake-compiler/issues/232#issuecomment-1904761619 as supported.

PR #236, opened 2024-01-28
set `required_rubygems_version` for native gems that specify the linux libc

flavorjones:
### Problem I'm trying to solve

Rubygems does not correctly recognize `-musl` or `-gnu` platform suffixes until v3.3.22.

### Solution

If rake-compiler is building a linux native gem that specifies a libc in its platform object, then add ">= 3.3.22" to the required_rubygems_version requirement.

https://github.com/rubygems/rubygems/blob/master/CHANGELOG.md#3322--2022-09-07

### Context

While working on musl support in the precompilation toolchain:

- https://github.com/rake-compiler/rake-compiler-dock/pull/111
- https://github.com/flavorjones/ruby-c-extensions-explained/pull/27
- https://github.com/sparklemotion/sqlite3-ruby/pull/442
- https://github.com/sparklemotion/nokogiri/pull/3111

I noticed that Ruby 3.0 is still shipping with Rubygems 3.2.33, which does not recognize these gem platforms. 

Specifying the rubygems requirement changes the error experienced by users during gem installation from:

> ERROR:  While executing gem ... (Gem::Exception)
>     Unable to find spec for #<Gem::NameTuple rcee_precompiled, 0.6.test.2024.0128.1724, aarch64-linux>

to:

> ERROR:  Error installing rcee_precompiled-0.6.test.2024.0128.1735-x86_64-linux-musl.gem:
>       rcee_precompiled-0.6.test.2024.0128.1735-x86_64-linux-musl requires RubyGems version >= 3.3.22.
> The current RubyGems version is 3.2.33. Try 'gem update --system' to update RubyGems itself.

/cc @deivid-rodriguez @segiddins

### DIFF
--- a/lib/rake/extensiontask.rb
+++ b/lib/rake/extensiontask.rb
@@ -278,6 +278,16 @@ Java extension should be preferred.
             "< #{ruby_api_version(sorted_ruby_versions.last).succ}.dev"
           ]
 
+          # set rubygems version constraints
+          if Gem::Version.new(Gem::VERSION) >= Gem::Version.new("3.3.22") &&
+             spec.platform.os == "linux" && !spec.platform.version.nil?
+            spec.required_rubygems_version = if spec.required_rubygems_version == Gem::Requirement.default
+              [">= 3.3.22"]
+            else
+              Gem::Requirement.new(gem_spec.required_rubygems_version, ">= 3.3.22")
+            end
+          end
+
           # clear the extensions defined in the specs
           spec.extensions.clear
 
--- a/spec/lib/rake/extensiontask_spec.rb
+++ b/spec/lib/rake/extensiontask_spec.rb
@@ -620,6 +620,93 @@ describe Rake::ExtensionTask do
         spec.metadata['allowed_push_host'].should eq 'http://test'
       end
 
+      it "should set required_rubygems_version when building a gem for `-linux-gnu` or `-linux-musl`" do
+        platforms = ["arm64-darwin", "arm-linux", "arm-linux-gnu", "arm-linux-musl"]
+        ruby_cc_versions = ["3.3.0"]
+        ENV["RUBY_CC_VERSION"] = ruby_cc_versions.join(":")
+
+        ruby_cc_versions.each do |ruby_cc_version|
+          platforms.each do |platform|
+            rbconf = "/rubies/#{ruby_cc_version}/rbconfig.rb"
+            allow_any_instance_of(Rake::CompilerConfig).to(
+              receive(:find)
+                .with(ruby_cc_version, platform)
+                .and_return(rbconf)
+            )
+          end
+        end
+
+        allow(Gem).to receive_message_chain(:configuration, :verbose=).and_return(true)
+
+        spec = Gem::Specification.new do |s|
+          s.name = 'my_gem'
+          s.platform = Gem::Platform::RUBY
+          s.extensions = ['ext/somegem/extconf.rb']
+        end
+
+        cross_specs = {} # platform => spec
+        Rake::ExtensionTask.new("extension_one", spec) do |ext|
+          ext.cross_platform = platforms
+          ext.cross_compile = true
+          ext.cross_compiling do |cross_spec|
+            cross_specs[cross_spec.platform.to_s] = cross_spec
+          end
+        end
+        platforms.each do |platform|
+          Rake::Task["native:my_gem:#{platform}"].execute
+        end
+
+        cross_specs["arm64-darwin"].required_rubygems_version.should eq(Gem::Requirement.default)
+        cross_specs["arm-linux"].required_rubygems_version.should eq(Gem::Requirement.default)
+
+        if Gem::Version.new(Gem::VERSION) >= Gem::Version.new("3.3.22")
+          expected_rubygems_version = Gem::Requirement.new([">= 3.3.22"])
+          cross_specs["arm-linux-gnu"].required_rubygems_version.should eq(expected_rubygems_version)
+          cross_specs["arm-linux-musl"].required_rubygems_version.should eq(expected_rubygems_version)
+        end
+      end
+
+      it "should append to required_rubygems_version if it's set" do
+        platform = "arm-linux-musl"
+        ruby_cc_versions = ["3.3.0"]
+        ENV["RUBY_CC_VERSION"] = ruby_cc_versions.join(":")
+
+        ruby_cc_versions.each do |ruby_cc_version|
+          rbconf = "/rubies/#{ruby_cc_version}/rbconfig.rb"
+          allow_any_instance_of(Rake::CompilerConfig).to(
+            receive(:find)
+              .with(ruby_cc_version, platform)
+              .and_return(rbconf)
+          )
+        end
+
+        allow(Gem).to receive_message_chain(:configuration, :verbose=).and_return(true)
+
+        spec = Gem::Specification.new do |s|
+          s.name = 'my_gem'
+          s.platform = Gem::Platform::RUBY
+          s.extensions = ['ext/somegem/extconf.rb']
+          s.required_rubygems_version = "!= 3.4.1" # keep this around
+        end
+
+        actual_cross_spec = nil
+        Rake::ExtensionTask.new("extension_one", spec) do |ext|
+          ext.cross_platform = [platform]
+          ext.cross_compile = true
+          ext.cross_compiling do |cross_spec|
+            actual_cross_spec = cross_spec
+          end
+        end
+        Rake::Task["native:my_gem:#{platform}"].execute
+
+        if Gem::Version.new(Gem::VERSION) < Gem::Version.new("3.3.22")
+          actual_cross_spec.required_rubygems_version.should eq(Gem::Requirement.new(["!= 3.4.1"]))
+        else
+          expected_rubygems_version = Gem::Requirement.new(["!= 3.4.1", ">= 3.3.22"])
+          actual_cross_spec.required_rubygems_version.should eq(expected_rubygems_version)
+        end
+      end
+
       after :each do
         ENV.delete('RUBY_CC_VERSION')
       end


segiddins (approved):

lib/rake/extensiontask.rb
```diff
@@ -278,6 +278,16 @@ def define_native_tasks(for_platform = nil, ruby_ver = RUBY_VERSION, callback =
             "< #{ruby_api_version(sorted_ruby_versions.last).succ}.dev"
           ]
 
+          # set rubygems version constraints
+          if spec.platform.os == "linux" && !spec.platform.version.nil?
+            if spec.required_rubygems_version == Gem::Requirement.default
+              spec.required_rubygems_version = [">= 3.3.22"]
+            else
+              spec.required_rubygems_version = Gem::Requirement.new(gem_spec.required_rubygems_version)
+              spec.required_rubygems_version.concat([">= 3.3.22"])
```
kou, line 287: Can we simplify this?

```suggestion
              spec.required_rubygems_version = spec.required_rubygems_version.requirements + [">= 3.3.22"]
```

spec/lib/rake/extensiontask_spec.rb
```diff
@@ -620,6 +620,102 @@
...
+          end
+        end
+
+        allow(Gem).to receive_message_chain(:configuration, :verbose=).and_return(true)
+
+        spec = Gem::Specification.new do |s|
+          s.name = 'my_gem'
+          s.platform = Gem::Platform::RUBY
+          s.extensions = ['ext/somegem/extconf.rb']
+          s.metadata['allowed_push_host'] = 'http://test'
```
kou, line 649: Can we remove this?

```suggestion
```

spec/lib/rake/extensiontask_spec.rb
```diff
@@ -620,6 +620,102 @@
         spec.metadata['allowed_push_host'].should eq 'http://test'
       end
 
+      it "should set required_rubygems_version when building a gem for `-linux-gnu` or `-linux-musl`" do
+        if Gem::Version.new(Gem::VERSION) < Gem::Version.new("3.3.22")
+          skip "rubygems 3.3.22+ only"
+        end
+
```
kou, line 627: Can we remove this and change expected value depending on RubyGems version instead?

```ruby
if Gem::Version.new(Gem::VERSION) < Gem::Version.new("3.3.22")
  cross_specs["arm-linux-gnu"].required_rubygems_version.should eq(Gem::Requirement.default)
  cross_specs["arm-linux-musl"].required_rubygems_version.should eq(Gem::Requirement.default)
else
  cross_specs["arm-linux-gnu"].required_rubygems_version.should eq(expected_rubygems_version)
  cross_specs["arm-linux-musl"].required_rubygems_version.should eq(expected_rubygems_version)
end
```

spec/lib/rake/extensiontask_spec.rb
```diff
@@ -620,6 +620,102 @@
...
+        cross_specs["arm-linux"].required_rubygems_version.should eq(Gem::Requirement.default)
+        cross_specs["arm-linux-gnu"].required_rubygems_version.should eq(expected_rubygems_version)
+        cross_specs["arm-linux-musl"].required_rubygems_version.should eq(expected_rubygems_version)
+      end
+
+      it "should append to required_rubygems_version if it's set" do
+        if Gem::Version.new(Gem::VERSION) < Gem::Version.new("3.3.22")
+          skip "rubygems 3.3.22+ only"
+        end
+
```
kou, line 676: ditto.

spec/lib/rake/extensiontask_spec.rb
```diff
@@ -620,6 +620,102 @@
...
+          end
+        end
+
+        allow(Gem).to receive_message_chain(:configuration, :verbose=).and_return(true)
+
+        spec = Gem::Specification.new do |s|
+          s.name = 'my_gem'
+          s.platform = Gem::Platform::RUBY
+          s.extensions = ['ext/somegem/extconf.rb']
+          s.metadata['allowed_push_host'] = 'http://test'
```
kou, line 698: ditto.

lib/rake/extensiontask.rb
flavorjones: `Gem::Requirements` does not support the `#+` method. See https://docs.ruby-lang.org/en//3.2/Gem/Requirement.html for supported methods. This is the simplest form I can think of.

deivid-rodriguez (approved):
:+1:

spec/lib/rake/extensiontask_spec.rb
flavorjones: Please note that for earlier versions of Rubygems, `-linux-gnu` is not supported, and `-linux-musl` _might_ be supported, depending on the versions. So I've updated this spec to assert only on what we care about:

```ruby
        cross_specs["arm64-darwin"].required_rubygems_version.should eq(Gem::Requirement.default)
        cross_specs["arm-linux"].required_rubygems_version.should eq(Gem::Requirement.default)

        if Gem::Version.new(Gem::VERSION) >= Gem::Version.new("3.3.22")
          expected_rubygems_version = Gem::Requirement.new([">= 3.3.22"])
          cross_specs["arm-linux-gnu"].required_rubygems_version.should eq(expected_rubygems_version)
          cross_specs["arm-linux-musl"].required_rubygems_version.should eq(expected_rubygems_version)
        end
```

and it does not assert on what we don't care about (the possible values of the unsupported platforms' specs). I hope this is OK.

spec/lib/rake/extensiontask_spec.rb
flavorjones: Similar to above: the platform `arm-linux-musl` _may_ be supported by rubygems (and so may end up as either `arm-linux` or `arm-linux-musl` once parsed), so the modified spec does not assert on the platform, and only asserts on the spec:

```ruby
        if Gem::Version.new(Gem::VERSION) < Gem::Version.new("3.3.22")
          actual_cross_spec.required_rubygems_version.should eq(Gem::Requirement.new(["!= 3.4.1"]))
        else
          actual_cross_spec.required_rubygems_version.should eq(expected_rubygems_version)
        end
```

I hope this is OK.

lib/rake/extensiontask.rb
kou: I think that `spec.required_rubygems_version.requirements` is an `Array` not a `Gem::Requirements`.

spec/lib/rake/extensiontask_spec.rb
```diff
@@ -620,6 +620,98 @@
...
+        end
+        platforms.each do |platform|
+          Rake::Task["native:my_gem:#{platform}"].execute
+        end
+
+        expected_rubygems_version = Gem::Requirement.new(["!= 3.4.1", ">= 3.3.22"])
+
+        if Gem::Version.new(Gem::VERSION) < Gem::Version.new("3.3.22")
+          actual_cross_spec.required_rubygems_version.should eq(Gem::Requirement.new(["!= 3.4.1"]))
+        else
```
kou, line 710: Could you reduce(?) scope?

```suggestion
        if Gem::Version.new(Gem::VERSION) < Gem::Version.new("3.3.22")
          actual_cross_spec.required_rubygems_version.should eq(Gem::Requirement.new(["!= 3.4.1"]))
        else
          expected_rubygems_version = Gem::Requirement.new(["!= 3.4.1", ">= 3.3.22"])
```

spec/lib/rake/extensiontask_spec.rb
kou: I see.

spec/lib/rake/extensiontask_spec.rb
kou: I see.

lib/rake/extensiontask.rb
flavorjones: I've changed this to

```ruby
spec.required_rubygems_version = Gem::Requirement.new(gem_spec.required_rubygems_version, ">= 3.3.22")
```

@kou What you suggest doesn't work. `requirements` is an array, but `Gem::Specification#required_rubygems_version=` does not accept it as an argument:

```ruby
spec = Gem::Specification.new do |s|
  s.name = 'my_gem'
  s.platform = Gem::Platform::RUBY
  s.extensions = ['ext/somegem/extconf.rb']
  s.required_rubygems_version = "!= 1.0.0"
end

spec.required_rubygems_version.requirements + [">= 3.3.22"]
# => [["!=", Gem::Version.new("1.0.0")], ">= 3.3.22"]

spec.required_rubygems_version = spec.required_rubygems_version.requirements + [">= 3.3.22"]
# /home/flavorjones/.rbenv/versions/3.2.3/lib/ruby/3.2.0/rubygems/requirement.rb:107:in `parse': Illformed requirement ["!="] (Gem::Requirement::BadRequirementError)
# 
#       raise BadRequirementError, "Illformed requirement [#{obj.inspect}]"
#             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
# 	from /home/flavorjones/.rbenv/versions/3.2.3/lib/ruby/3.2.0/rubygems/requirement.rb:139:in `block in initialize'
# 	from /home/flavorjones/.rbenv/versions/3.2.3/lib/ruby/3.2.0/rubygems/requirement.rb:139:in `map!'
# 	from /home/flavorjones/.rbenv/versions/3.2.3/lib/ruby/3.2.0/rubygems/requirement.rb:139:in `initialize'
# 	from /home/flavorjones/.rbenv/versions/3.2.3/lib/ruby/3.2.0/rubygems/requirement.rb:64:in `new'
# 	from /home/flavorjones/.rbenv/versions/3.2.3/lib/ruby/3.2.0/rubygems/requirement.rb:64:in `create'
# 	from /home/flavorjones/.rbenv/versions/3.2.3/lib/ruby/3.2.0/rubygems/specification.rb:702:in `required_rubygems_version='
# 	from /home/flavorjones/bar.rb:15:in `<main>'
```

flavorjones: I think that `Gem::Requirement` could be improved in two ways that would make this feature simpler to implement:

1. `#dup` should make a deep copy of `@requirements`. It does not, and so I am calling `Gem::Requirement.new(gem_spec.required_rubygems_version)` to duplicate it.
2. `#concat` could return `self`. It does not, and so I am doing this across two statements.

If those things were fixed, this could become

```ruby
spec.required_rubygems_version = spec.required_rubygems_version.dup.concat(">= 3.3.22")
```

spec/lib/rake/extensiontask_spec.rb
flavorjones: I'm not sure what you mean. I think you mean make the test smaller or simpler?

I've tried to simplify this test slightly by removing the array of platforms. Is that helpful? Can you explain more what you mean?

flavorjones: Ah, if you mean "move `expected_rubygems_version` after the `else` then I have done that.

flavorjones: Done.

lib/rake/extensiontask.rb
kou: Oh. I see.

kou:
Thanks!

flavorjones:
See https://github.com/rubygems/rubygems/pull/7439 which proposes changing Gem::Requirement and Gem::Specification to deep-copy requirements.

PR #239, opened 2024-09-04
Update mingw website

evanfarrar:
The project moved off of sourceforge a long time ago. Unfortunately, I can't find the files that the readme references, so maybe this whole section should be removed since it is sort of an area a user will have to research today if they want to avoid rake-compile-dock on OSX.


kou:
How about using https://www.mingw-w64.org/ instead?

evanfarrar:
Ah, this was my confusion, you recommend mingw-w64, which is a fork still hosted on sourceforge, where as mingw (which now supports 64 bit and was formerly know as mingw-w32) has moved to OSDN.

PR #240, opened 2024-10-04
Don't generate additional rake tasks when files don't exist

y-yagi:
If files is retrieved from Git and the task is executed after the files are deleted but before they are committed to Git, an error occurs because the files do not exist.

Getting files from Git is a common pattern because it's default of newgems by generated by Bundler.
https://github.com/rubygems/rubygems/blob/ad394cadc73f2239b907e6cd02a0dfae2ef65999/bundler/lib/bundler/templates/newgem/newgem.gemspec.tt#L31-L35

I would like to avoid an error in this situation, so I fixed not to create rake tasks for compile if files don't exist.

### DIFF
--- a/lib/rake/extensiontask.rb
+++ b/lib/rake/extensiontask.rb
@@ -86,8 +86,8 @@ module Rake
       lib_binary_path = "#{lib_path}/#{File.basename(binary(platf))}"
 
       files.each do |gem_file|
-        # ignore directories and the binary extension
-        next if File.directory?(gem_file) || gem_file == lib_binary_path
+        # ignore directories, the binary extension and non-existent files
+        next if File.directory?(gem_file) || gem_file == lib_binary_path || !File.exist?(gem_file)
         stage_file = "#{stage_path}/#{gem_file}"
 
         # copy each file from base to stage directory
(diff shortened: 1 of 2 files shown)


kou:
Thanks. I've release a new version.

y-yagi:
Thanks for your quick response!

larskanis:
In fxruby I have a regression due to this change: https://github.com/larskanis/fxruby/actions/runs/11837782836/job/33043780149#step:9:12

This is because the file `fox16/kwargs.rb` is generated per `Rakefile` task, and the task is no longer triggered by rake-compiler-1.2.8.

I'm not sure if this is a intended use case and if it legitimates a rollback - I can probably use a different way to generate the file. But it is at least a real world regression.

kou:
@y-yagi Do you have any opinion for this case?

y-yagi:
@kou I'm not sure it was a supported behavior, but I agree that this made a regression. 
I was thinking about how to fix it, but I couldn't find a good solution. So I think we should revert this. Can I make a PR for it?

kou:
OK. I'll think about this carefully this later. Please wait for a while.

kou:
I considered this case.

Let's revert this.

Specifying nonexistent files (that aren't generated automatically) to `spec.files` is a bug of users. `git ls-files` may be convenient but committing deleted files before `rake compile` is users' responsibility not rake-compiler's responsibility.

PR #247, opened 2025-04-11
javaextensiontask.rb - check `javac` path using `JAVA_HOME`

MSP-Greg:
JRuby-head is using JDK 21.  This is not the default on GHA.  `setup-ruby` added code to change `JAVA_HOME`, assuming it would trigger the correct JDK being used.

Puma tests against jruby-head, and it was failing when rake compile ran.  After reviewing the code, `javac` is being called without a path.

Added code to check if `File.join(ENV["JAVA_HOME"], "bin", "javac")` exists.  If it does, use the full path, otherwise fallback to current behavior.

I changed the Gemfile in Puma to use the PR branch, and the job correctly compiles.

### DIFF
--- a/lib/rake/javaextensiontask.rb
+++ b/lib/rake/javaextensiontask.rb
@@ -108,8 +108,15 @@ execute the Rake compilation task using the JRuby interpreter.
         EOF
         warn_once(not_jruby_compile_msg) unless defined?(JRUBY_VERSION)
 
+        java_home = ENV["JAVA_HOME"]
+        if java_home
+          javac_path = File.join(java_home, "bin", "javac")
+          javac_path = nil unless File.exist?(javac_path)
+        end
+        javac_path ||= "javac" 
+
         javac_command_line = [
-          "javac",
+          javac_path,
           *java_target_args,
           java_lint_arg,
           "-d", tmp_path,
@@ -306,7 +313,7 @@ execute the Rake compilation task using the JRuby interpreter.
 
     def release_flag_supported?
       return true unless RUBY_PLATFORM =~ /java/
-      
+
       Gem::Version.new(Java::java.lang.System.getProperty('java.version')) >= Gem::Version.new("9")
     end
   end


MSP-Greg:
@headius

Sorry for the ping.  You and Pavel have the two most recent commits here.  I'm somewhat java challenged.  Does this look ok?

JFYI, started with Puma failing on jruby-head...

MSP-Greg:
@headius

Sorry for the ping.  You and Pavel have the two most recent commits here.  I'm somewhat java challenged.  Does this look ok?

JFYI, started with Puma failing on jruby-head...

lib/rake/javaextensiontask.rb
```diff
@@ -108,8 +108,11 @@ def define_compile_tasks(for_platform = nil, ruby_ver = RUBY_VERSION)
         EOF
         warn_once(not_jruby_compile_msg) unless defined?(JRUBY_VERSION)
 
+        javac_path = File.exist?(t = File.join(ENV["JAVA_HOME"], "bin", "javac")) ?
```
kou, line 111: It seems that `ENV["JAVA_HOME"]` may not exist:

https://github.com/rake-compiler/rake-compiler/blob/150c8e359748a4e7582daea6ecbe7ecb063d3086/lib/rake/javaextensiontask.rb#L267-L268

Can we reuse the checked `ENV["JAVA_HOME"]` value?

MSP-Greg: Not sure if `ENV["JAVA_HOME"]` is used anywhere?

Can we use:
```ruby
ENV.fetch("JAVA_HOME", "/")
```

MSP-Greg: I just pushed a change to using `ENV.fetch`.  Used `"/"` for the default so it's an absolute path...

kou: Oh, sorry. It's `JRUBY_HOME` not `JAVE_HOME`...

lib/rake/javaextensiontask.rb
```diff
@@ -108,8 +108,11 @@ def define_compile_tasks(for_platform = nil, ruby_ver = RUBY_VERSION)
         EOF
         warn_once(not_jruby_compile_msg) unless defined?(JRUBY_VERSION)
 
+        javac_path = File.exist?(t = File.join(ENV.fetch("JAVA_HOME", "/"), "bin", "javac")) ?
+          t : "javac"
```
kou, line 112: Using `/` as the default `JAVA_HOME` will not work when `javac` exists in both of `/usr/bin/javac` and `/usr/local/bin/javac`, and `PATH` is `/usr/local/bin:/usr/bin:/bin`.

We may need to check `ENV["JAVA_HOME"]` explicitly:

```suggestion
        java_home = ENV["JAVA_HOME"]
        if java_home
          javac_path = File.join(java_home, "bin", "javac")
          javac_path = nil unless File.exist?(javac_path)
       end
       javac_path ||= "javac" 
```

MSP-Greg: @kou

Thanks for the review.  I'm not sure about your example, it seems a rather extreme and odd edge case.  Also, I'm not familiar with all of the *nix distributions.

`"/"` was a bad choice for default, as `File.join('/', 'bin', 'javac')"` and `File.join('', 'bin', 'javac')` both equal `"/bin/javac"`.

The previous code worked correctly when used with GHA.  But your changes also work.

Updated with your suggested code.

MSP-Greg:
Some background on this issue.

At present, GitHub Actions (GHA) runners provide at least three Java versions (11, 17, 21).  Some provide version 8.  Version 17 is the default.  `ENV` variables exist for determining the location of each version.

JRuby 10/head requires version 21.  [`setup-ruby`](https://github.com/ruby/setup-ruby/pull/721) made changes for this by changing the value of `ENV['JAVA_HOME']`.

This PR finishes the work by using `ENV['JAVA_HOME']` to determine the path for the correct version of `javac` to run for the compile process.

lib/rake/javaextensiontask.rb
kou: Oh, sorry. I considered the following case:

* `ENV["PATH"]` is `"/usr/local/bin:/usr/bin`
* `ENV["JAVA_HOME"]` isn't set
* `/usr/local/bin/javac` exists
* `/usr/bin/javac` exists

In the case, `javac` uses `/usr/local/bin/javac` but `File.join(ENV.fetch("JAVA_HOME", "/"), "bin", "javac")` uses `/usr/bin/javac` not `/usr/local/bin/javac`.

kou:
@headius I'll merge this and release a new version.

If you have any problem with this, please share it.

lib/rake/javaextensiontask.rb
MSP-Greg: Sorry, I need to try this.

If `ENV["JAVA_HOME"]` isn't set, then `ENV.fetch("JAVA_HOME", "/"), "bin", "javac")` yields `/bin/javac`, I thought `File.exist?("/bin/javac")` would be false, and then `javac` would be used, and the file would be resolved by `PATH`?

Again, I need to try it...

eregon:
Thanks for this PR.

It works fine on UNIX but not Windows:
https://github.com/eregon/actions-shell/actions/runs/20195116809/job/57977991482#step:4:369
```
ENV["JAVA_HOME"]
"C:\\hostedtoolcache\\windows\\Java_Temurin-Hotspot_jdk\\21.0.9-10.0\\x64"

File.join(ENV["JAVA_HOME"], "bin", "javac")
"C:\\hostedtoolcache\\windows\\Java_Temurin-Hotspot_jdk\\21.0.9-10.0\\x64/bin/javac"

File.exist?(File.join(ENV["JAVA_HOME"], "bin", "javac"))
false
```

I'm unsure if it's a bug of `File.join` on JRuby on Windows (the code does `File.join(java_home, "bin", "javac")` which intuitively should work, either use only `\` or convert everything to `/` but a mix seems quite broken).
cc @headius

eregon:
It seems `File.join` does have those "broken" semantics even on CRuby :/
https://github.com/eregon/actions-shell/actions/runs/20195181454/job/57978154719#step:4:798
```
JAVA_HOME = "C:\\hostedtoolcache\\windows\\Java_Temurin-Hotspot_jdk\\21.0.9-10.0\\x64"
File.join(JAVA_HOME, "bin", "javac")
"C:\\hostedtoolcache\\windows\\Java_Temurin-Hotspot_jdk\\21.0.9-10.0\\x64/bin/javac"
```

So I guess we should just hack it with
```ruby
File.join(java_home.tr('\\', '/'), "bin", "javac")
```
then, or seeing that File.join is beyond useless here, just:
```ruby
"#{java_home.tr('\\', '/')}/bin/javac"
```

eregon:
Oof even that doesn't work, Windows is truly cursed IMO:
https://github.com/eregon/actions-shell/actions/runs/20195353255/job/57978552583#step:4:377
```
"#{JAVA_HOME.tr('\\', '/')}/bin/javac"
"C:/hostedtoolcache/windows/Java_Temurin-Hotspot_jdk/21.0.9-10.0/x64/bin/javac"

File.exist?("#{JAVA_HOME.tr('\\', '/')}/bin/javac")
false
```
At this point I'll let others investigate/fix it.

eregon:
Ah, it's my mistake on Windows there is no bin/javac, it's bin/javac.exe.
Then all of them work:
```
 File.join(JAVA_HOME, "bin", "javac.exe")
"C:\\hostedtoolcache\\windows\\Java_Temurin-Hotspot_jdk\\21.0.9-10.0\\x64/bin/javac.exe"

File.exist?(File.join(JAVA_HOME, "bin", "javac.exe"))
true

"#{JAVA_HOME.tr('\\', '/')}/bin/javac.exe"
"C:/hostedtoolcache/windows/Java_Temurin-Hotspot_jdk/21.0.9-10.0/x64/bin/javac.exe"

File.exist?("#{JAVA_HOME.tr('\\', '/')}/bin/javac.exe")
true

"#{JAVA_HOME}\\bin\\javac.exe"
"C:\\hostedtoolcache\\windows\\Java_Temurin-Hotspot_jdk\\21.0.9-10.0\\x64\\bin\\javac.exe"

File.exist?("#{JAVA_HOME}\\bin\\javac.exe")
true
```
Sanity has returned.

eregon:
@kou so there is https://github.com/rake-compiler/rake-compiler/commit/db38e70ffbc0ef382ff6f26775d2a240fd8d4474 but https://rubygems.org/gems/rake-compiler only has 1.3.0 not 1.3.1, so it seems you forgot to push the gem or it didn't work?
Could you push 1.3.1 (or 1.3.2 if 1.3.1 is not possible?)

That would also explain why CI failed on https://github.com/ruby/strscan/pull/155

And now I found the discussion I recalled: https://github.com/ruby/strscan/pull/151
I had hope @headius would look into this before we run into this again.

eregon:
@headius IMO this is messy enough that I think it would be better to add `JAVA_HOME/bin` to `PATH` in setup-ruby.
I don't see a disadvantage to do so, setup-java would do the same thing and it's likely to work poorly to have `java` in PATH different than `JAVA_HOME/bin/java` (e.g. javac uses PATH, Maven uses JAVA_HOME, a wild mix).

eregon:
I went ahead and did a PR to do that so we can't run into such issues again: https://github.com/ruby/setup-ruby/pull/836

kou:
Oh, sorry. I didn't notice it.

I've published 1.3.1.

PR #257, opened 2026-09-18
[Content Addressable] Add content addressable / single Ruby ABI gem support in rake-compiler

OughtPuts:
https://github.com/rake-compiler/rake-compiler/issues/258

## TLDR

Implements some small changes in order to support the build and packaging of single ABI gems, which are content addressed. 

## Description

This PR:

1. Adjusts the `spec.required_ruby_version` to pessimistic version syntax in the case of a single Ruby ABI to align with the expectations in the downstream `rubygems` changes for content addressable gems.

2. Makes the `ruby_abi` attribute available to `Gem::PackageTask.new` so that this eventually can be provided to `Gem::Package.build`. 

3. Sets up a sub directory scoped by `ruby_abi` for the eventual gem build to be stored inside, so that on line 342 of `extention_task`, when we check the directory for existence / staleness to decide whether or not the task needs to run, we will not get a false negative from having stored the multi-abi build in the root directory earlier in the flow. 

## Testing

Tests added to `extensiontask_spec`. 

## Top Hat

This has been top-hatted in the flow as a whole, see the [cibuildgem CA support PR](https://github.com/Shopify/cibuildgem/pull/62#issue-5314244278) for more information.


lib/rake/extensiontask.rb
```diff
@@ -273,10 +273,21 @@ def define_native_tasks(for_platform = nil, ruby_ver = RUBY_VERSION, callback =
...
             ruby_version.split(".").collect(&:to_i)
           end
-          spec.required_ruby_version = [
-            ">= #{ruby_api_version(sorted_ruby_versions.first)}",
-            "< #{ruby_api_version(sorted_ruby_versions.last).succ}.dev"
-          ]
+
+          # introduce pessimistic version requirement syntax for skinny packaged gems
+          ruby_abi =
+            if sorted_ruby_versions.one?
```
OughtPuts, line 279: I've just thought that this will change the syntax for a gem that is not designed to be content-addressable but just happens to support only one ruby ABI so I need to look at this again - I think we need an additional check here as well as `sorted_ruby_versions.one?`

Edouard-chin: Isn't a gem that only supports one ABI the same as a content-addressable gem ?

kou: Could you add `ExtensionTask#content_addressable=` instead? If it's `true`, can we generate content addressable gem per Ruby version? For example, if we have `3.3.0`, `3.4.0` and `4.0.0`, can we generate 3 content addressable gems?

As I mentioned at https://github.com/ruby/rubygems/pull/9773#issuecomment-5690252169, I feel that controlling content addressable or not by `ruby_abi` is strange. Can we use an explicit flag whether we use content addressable or not instead of `ruby_abi`?

kou:
> https://github.com/shop/issues-rails-infrastructure/issues/1746

I can't see this.

OughtPuts:
Putting this in draft temporarily until we are aligned on any changes to the `--ruby_abi` build flag.

OughtPuts:
> > [shop/issues-rails-infrastructure#1746](https://github.com/shop/issues-rails-infrastructure/issues/1746)
> 
> I can't see this.

Now moved across to this issue - apologies! https://github.com/rake-compiler/rake-compiler/issues/258

kou:
Thanks!

lib/rake/extensiontask.rb
OughtPuts: @Edouard-chin yes, ignore me, I lost track of the fact we were dealing with only compiling gems at this point so the only options are multi ABI or single ABI. 

@kou yes, no probs!

